PR #2, opened 2020-10-22
code_800393C0

Kelebek1:
func_8003DB5C and func_8003D080 not matching due to a conflicting argument type (s32 vs f32), and func_8003C0F0 also has issues with calling spawn_players_2p_battle with a float arg as well.


MegaMech:
lgtm! A couple comments:

Should we name the file? Maybe spawning.c or any suitable name is fine.
Also, in asm/non_matchings/code_800393C0/ you split each function into its own asm file. Is it necessary to have an asm file for all the functions you matched? Perhaps just the functions that aren't matching?

Kelebek1:
I don't know what to name it, that's up to you guys. I didn't split the functions manually, asm-splitter.py did it, I assume that's what most code files will do? Yeah not all are needed, and they shouldn't be needed once matched, but there's no script to delete unused asm like OOT/MM has so I just left them. With the linker script editing they're not used anyways.

ProjectRevoTPP (approved):

MegaMech:
Thank you for the submission. We really appreciate it. This PR has been transferred to the private repository and will be in the next refresh. If we did something wrong in that process and you were not properly credited for your efforts just let us know and it will be fixed.

Thank you.

PR #4, opened 2020-11-06
added --no-builtin-rules to makeflags to improve build speed

farisawan-2000:


MegaMech:
Closed because the change is in the private repo.

PR #7, opened 2021-03-01
Rename known offsets

duncanspumpkin:
~~Gets `init_controllers` in a conformant state and also~~ renames a couple of offsets as per SM64 as they are doing the exact same job.


MegaMech:
lgtm, thanks!

PR #35, opened 2021-10-11
Match `func_80000D3C` (renamed) and `thread5_game_logic`

iProgramMC:
Also attempted to match `init_game`, but some registers wouldn't match


include/segments.h
```diff
@@ -4,5 +4,6 @@
 #define SEG_MAIN     0x80000400
 #define SEG_80280000 0x80280000
 #define SEG_8028DF00 0x8028DF00
+#define SEG_802BA370 0x802BA370
```
MegaMech, line 7: What's located at `802BA370`?

src/code_80281C40.c
```diff
@@ -52,7 +52,7 @@ void func_80281D00(void) {
     func_802A53A4();
     func_802A3E3C();
     if (D_80287550 != 0xFFFF) {
-        func_80000D3C(0);
+        clear_framebuffer(0);
```
MegaMech, line 55: Does this match the func name from sm64 if it has the same func?

src/main.c
```diff
@@ -129,31 +129,30 @@ void create_thread(OSThread *thread, OSId id, void (*entry)(void *), void *arg,
 
 void main_func(void) {
     osInitialize();
-    create_thread(&D_801524C0, 1, thread1_idle, NULL, &D_80154670, 100);
-    osStartThread(&D_801524C0);
+	//! Why is this using the gVideoThread as idle stack?
+    create_thread(&gIdleThread, 1, thread1_idle, NULL, &gVideoThread, 100);
```
MegaMech, line 133: Is it possible they chose gVideoThread due to the below size requirements? Looks like the stack size of the idlethread is set to the size of the gVideoThread. I tried checking the original asm but it likely no longer exists. Could double check that `&gVideoThread` set as the stack wasn't a mistake on our part or the wrong name.
```
Similarly, the stack is initialized to the location pointed to by the sp argument. Note that stacks grow backwards (towards smaller addresses) with the MIPS compiler tools. The memory region assigned to the stack must be of least size OS_MIN_STACKSIZE in bytes. The maximum size of the stack must be the sum of the stack frame sizes of its maximum procedure call depth plus 16. Stacks must also be 64-bit aligned, so they should be declared as 'long long int' and type-casted when passed to this function. Use the stacktool program to help compute stack sizes.
```

src/main.c
```diff
@@ -129,31 +129,30 @@ void create_thread(OSThread *thread, OSId id, void (*entry)(void *), void *arg,
...
     }
     osViBlack(TRUE);
     osViSetSpecialFeatures(OS_VI_GAMMA_OFF);
     osCreatePiManager(OS_PRIORITY_PIMGR, &D_8015F460, &D_8015F3E0, 0x20);
     D_8015011C = (s16) osResetType;
     create_debug_thread();
     start_debug_thread();
-    create_thread(&D_80154670, 3, &thread3_video, arg0, &D_80156820, 100);
-    osStartThread(&D_80154670);
+    create_thread(&gVideoThread, 3, &thread3_video, arg0, &D_80156820, 100);
```
MegaMech, line 150: I don't suppose you could also rename D_80156820 to:
`gVideoThreadStack`

src/main.c
```diff
@@ -463,42 +459,46 @@ GLOBAL_ASM("asm/non_matchings/main/dma_copy.s")
...
     temp_a1 = gPrevLoadedAddress;
-    sp38 = temp_a1;
     sp40 = ALIGN16(sp2C[1]);
     mio0decode(sp2C, temp_a1);
     set_segment_base_addr(0xD, temp_a1);
-    temp_t7 = sp40 + gPrevLoadedAddress;
-    gPrevLoadedAddress = temp_t7;
-    D_8015F734 = temp_t7;
+    
+    gPrevLoadedAddress = gPrevLoadedAddress + sp40;
```
MegaMech, line 492:
```
gPrevLoadedAddress += sp40;
```

src/main.c
```diff
@@ -463,42 +459,46 @@ GLOBAL_ASM("asm/non_matchings/main/dma_copy.s")
...
+	
+	
+	//need to match this
+	
     temp_a1 = gPrevLoadedAddress;
-    sp38 = temp_a1;
     sp40 = ALIGN16(sp2C[1]);
     mio0decode(sp2C, temp_a1);
     set_segment_base_addr(0xD, temp_a1);
-    temp_t7 = sp40 + gPrevLoadedAddress;
```
MegaMech, line 491: I think the issue with matching here is likely temp_t7. It probably needs to disappear and instead it adds back and forth between either sp40 or gPrevLoadedAddress.
Ex:
```
gPrevLoadedAddress + sp40;
```

src/main.c
```diff
@@ -1211,7 +1215,7 @@ void thread5_game_logic(s32 arg0) {
     func_800C5CB8();
 loop_3:
```
MegaMech, line 1216: Try while(1) here since there doesn't appear to be a condition for the goto statement.

MegaMech:
This should be a more helpful idea to try than my last idea:
```
    //temp_a1 = gPrevLoadedAddress;
    sp40 = ALIGN16(sp2C[1]);
    mio0decode(sp2C, gPrevLoadedAddress);
    set_segment_base_addr(0xD, gPrevLoadedAddress);
```

iProgramMC:
So I merged MegaMech and CoderStig's changes, weirdly CI still fails

PR #56, opened 2021-11-13
[kart] Variety of changes related to kart hop physics

Tmcg2:
Gave some Player struct members proper names, mostly related to kart hop physics.

Renamed a couple of tables in kart_attributes to make their purpose and nature more obvious

Matched func_8002AAC0 with its assembly representation
Matched func_800225CC with its assembly representation


MegaMech:
Hi Tyler,

Thanks for the contribution! I really appreciate it.

Just a friendly reminder that contributors cannot use mk64 leaked source code.

For comments please try to construct them in this style:
```
/**
  * Multi-line comment. This style for documenting functions
  *
 **/

// Single line comment ex. before a loop or if.
```
Some of your comments take-up many lines. Your style is fine for that for now, but later on when that formatting will probably be changed. However, any documentation is good documentation.

Is KartId actually characterId? I would prefer the word character if kartId means ex. Yoshi.

Tmcg2:
No worries, I was not even aware that any source code had leaked prior to taking interest in the project. I have not seen nor pursued leaked source code for anything N64 related.

I'm on the fence about kartId. I did originally call it characterId but swapped to kartId due to being focused on the kart hop physics stuff. Looking at it now though the player->unk254 member is used in a lot of other files, most of which are probably not focused on kart physics stuff. I can swap it to characterId if that is desired, I'm not married to the current name.

I do not believe that playerId is accurate as modifications to the gKartHopInitalVelocity table only affect specific characters/karts.

Please pardon my verbosity, I tend to be a little rambly. I'll adjust the comment blocks to fit the expected style.

Do you prefer to have multiple commits per PR or should I squash any extra changes into a single commit?

MegaMech:
The spawn player func looks like this:
```
                        Player pointer,  start order or playerId,    start coordinates,                  unk,      characterModel,    type<player, ai>
spawn_player(D_800DC4FC, 0, arg0[D_80165270[0]], arg1[D_80165270[0]], arg2, 32768.0f, D_800E86A8[0], -0x1000);
```
The characterModel arg sets 254.
254(kartId) is set using func_80039F44 which randomly sets AI to characters/karts avoiding already selected human player characters.

D_800E86A8[4] is selectedCharacters. (one byte for each player)

playerId is definitely not correct for this scenario. characterId works but I can see why kartId might work better. The issue is that this member also chooses the character and not just kart physics. kartId seems too ambiguous to me, but others may feel differently.

PR #57, opened 2021-11-16
Found the variable that tracks the CC type selected.

Tmcg2:
Added some comments for the possible values of Mode and CC
selection


MegaMech:
Rename `gCcSelection` to `gCCSelection`
Engine displacement is measured using cubic centimetres. As such, it's two words and both letters should be capitalized (Plus this makes it more clear that it refers to CC).

Everything looks good. You'll need to refresh your branch or just solve the conflict. Same for your other PR. And then both can be merged.

Consider observing some sm64 comments. Generally, we want what a function is for/does while avoiding comments such as "if that then minus that does that." Something more like "Object moves in Y direction until out of range" kind of thing. The decomp is still in its early stages. So I won't be too picky, but just mentioning it for the future.

I think these two files are good examples for comments:
https://github.com/n64decomp/sm64/blob/master/src/game/behaviors/blue_fish.inc.c
https://github.com/n64decomp/sm64/blob/master/src/game/game_init.c

Also, if you like, you can join the decomp Discord: discord.gg/DuYH3Fh

Thanks!

PR #60, opened 2021-11-20
Jump Table Related Disassembly

Tmcg2:
The intent of this document is the explain the structure and content of this pull request in a more digestable fashion.
The pull request is broken up into 3 distinct parts, each comprised of several commits.
The intent is that within any given part, the commits may be accepted/rejected individually.
While later parts do have dependencies on earlier parts, the hope that rejecting any 1 commit has as little impact on the pull request overall as possible.

# Part 1: Identifying Jump Tables

This is a set of 24 commits, 1 commit per folder in ```asm/nonmatchings/``` that had at least 1 modified file.
Each file changed had an entry that was a jump table that was mislabelled as data.
That is to say, entries that were marked as data (D_\*) were changed to be marked as jump tables (jpt_\*).
These labels were defined in .roadata files (namely ```data_0DD0A0_3.s```) found in ```data/```, and as such needed to be updated in line with the changes to the assembly files.
By themselves, these changes do nothing and keep the build output identical to the base ROM.
However, they are a necessary pre-requisite for...

# Part 2: Disassembly of Functions that use Jump Tables

This is a set of 20 commits, 1 commit per file in ```src/``` that was modified.
Throughout several ```.c``` files in ```src/``` there were functions that had no disassembly due to ```mips_to_c``` being unable to find a jump table.
With the changes to label names from Part 1 however, ```mips_to_c``` could be told where to find the jump tables and create an initial disassembly of the function.
Each function generated this way was made with the following command:

```bash
mips_to_c.py --no-casts --rodata data/data_* -- asm/non_matchings/$1/$2.s;
```

Where the first argument ($1) was the name of a folder in ```asm/nonmatchings``` and the second argument ($2) was the name of a function (so, filenames minus the ```.s```).
Each function made this way was then added to the appropriate ```.c``` folder in ```src/```.
There are 2 major excpetions to this pattern: ```code_8028DF00.s``` and ```code_8008C1D0.s```.
While there were D_* to jpt_* changes made in those files, there was no easy way to create disasseblies of just the functions modified in this way.
Doing so would require breaking up the assembly file into several smaller parts in order to keep the build output identical to the base ROM.
But if an assembly file has to broken up at all, why not break it apart entirely?
Hence...

# Part 3: Breakup of Large Assembly Files

This is a set of 4 commits, 1 commit per large assembly file broken up.
There were 4 folders in ```asm/nonmatchings``` that contained large, monolithic ```.s``` files.
This made them stick out, as every other folder contains several small files, each dedicated to 1 function.
So, each large assembly file was broken up into individual function files.
Then, using the same command used in Part 2, each function file was disassembled and then placed into the appropriate ```.c``` file.
Of the 4 files broken up this way, only 2 have any jump table related changes: ```code_8008C1D0.s``` and ```code_8028DF00.s```.
Some functions disassembled in this way had no labels in the original ```.s``` file.
Its unclear if this indicates that the function is unused, unnamed, or was just an error in the original byte-code to assembly translation.
Care was taken to identify these types of functions in the disassembly added to the corresponding ```.c``` file, but its possible that some may have been missed.


Tmcg2:
I am lurking in the Discord if anyone wishes to ping me (Taggerung) there

MegaMech:
Thanks for the PR!

I'm going to have a deeper dive into this later. Here's a few initial thoughts:

The reason some asm files weren't split up properly was because I had already done some initial experimentation in that file. So I just lazily dumped all the asm into a single file. The splitter didn't grab these because I didn't put in the work to make it do that.

Just a heads up I have a PR coming for 8028DF00 that matches a bunch of funcs and may or may not split the remaining files (depending how lazy I am). So you may want to ignore that file for now. 
 
Also, please run M2C using the --context ctx.c command. I can provide you with a context file if you like. It's a bit of a rabbit hole to generate one.

MegaMech:
Reviewed. Looks good to me if we can resolve the below:

Would it be easy for you to re-do the m2c with a context file? I think it's worth it as the output is better.

MegaMech:
> Some functions disassembled in this way had no labels in the original
In answer to this question. Functions that are not referenced do not receive labels. So yes. Some func files will have extra unused funcs in them.

Tmcg2:
> Would it be easy for you to re-do the m2c with a context file? I think it's worth it as the output is better.

Sure, I can do that. I see there's a ```ctx_includes.c``` in repo, is that the appropriate file or would I want to make my own?

> Just a heads up I have a PR coming for 8028DF00 that matches a bunch of funcs and may or may not split the remaining files (depending how lazy I am). So you may want to ignore that file for now.

I was aware of that, I think you mentioned it in Discord sometime in the past few days. I'll drop commits related to that file as I handle the context file stuff for mips_to_c

MegaMech:
As long as the headers in ctx_includes are up to date. (I believe they are) Then yes.

In tools/ there's a file called `m2ctx` (.py) if you run that with ctx includes it should work. You may need to put it in the root directly mk64/ for it to work. Then you'll need to delete all the garbage error lines that aren't usable. (Or modify the script ot not generate them) if something doesn't work at all then these might need editing:
```
CPP_FLAGS = [
    "-Iinclude",
    "-Isrc",
    "-D_LANGUAGE_C",
    "-DF3DEX_GBI",
    "-D_MIPS_SZLONG=32",
    "-DSCRIPT(...)={}"
    "-D__attribute__(...)=",
    "-D__asm__(...)=",
    "-ffreestanding",
]
```
I thought I updated these to be correct. But I don't have time atm to double check.

PR #67, opened 2021-11-25
Give names to some Grand Prix related variables

Tmcg2:
I'm open to suggestions on the variable names since, as they stand now, they're a little on the long side


data/data_0DD0A0_3.s
```diff
@@ -5380,13 +5380,13 @@ glabel D_800F0B10
 .asciiz "    th"
 
 .balign 4
-glabel D_800F0B18
+glabel gGrandPrix1stPlacePointReward
```
MegaMech, line 5383: gGrandPrix1stPlacePointReward

How about->
gGrandPrix1stPlaceReward

I'm wondering if GrandPrix should be abreviated as:
gGP...

MegaMech (commented):
lgtm.

It would be nice to find a shorter version of gGrandPrixCurrentRaceStandings.
Perhaps GP instead of grandprix? But I dunno. At least it's clear what currentRaceStandings is.
gGPCurrentRaceStandings.

Are you sure this uses char ids? I looked at ram and it seemed slightly different than that. Perhaps player index? But even that I'm unsure about.

Tmcg2:
> Are you sure this uses char ids? I looked at ram and it seemed slightly different than that. Perhaps player index? But even that I'm unsure about.
Yes I'm pretty certain about that

8018D9C8 is a list of 8 s8's. If you complete a GP race and watch those values during the point allotment its pretty clear that index 0 is Mario's points, index 1 is Luigi's points, etc.

8018D9D0 is also a list of 8 s8's. However this time the contents are character ID's. So index 0 is the ID of the character in 1st place, index 1 is 2nd place, etc.

I've added some defines for the different character IDs so that there's a reference for anyone looking for them in the future

MegaMech:
Looks great. Yea that's much better. My only complaint now is this:
`gCharacterGPPoints`
How about:
`gGPPlayerPoints`
I think that's more consistent to put GP at the beginning.

Tmcg2:
I can move GP to the front, but I'm inclined to leave it as ```Character``` rather than ```Player```. The list is indexed by character ID rather than player ID and (to me at least) having ```Character``` in the name makes that more obvious.

PR #69, opened 2021-11-28
Add a structure for accessing save data

Tmcg2:
I'm very open to suggestions for names and such because in its current state the structure is super ugly.

Save data is saved as one big block starting at 8018EB90, going for 256 (0x200) bytes total 
Time Trial data and Grand Prix results are stored in this block
It is broken up into 2 major section.
The first section the course-by-course set of Time Trial records. Each course record is comprised of:
- The top 5 3-lap records
- The 1-lap record
- 6 bytes of unknown purpose
  - It seems like the first byte may be an indicator of whether or not a ghost is available for the course or not
The order of the courses is the same as that found in gCupTrackOrder, as opposed to the order found in the COURSES enum in common_structs.h,
That is to say, the order is based on what might be called the "courseIndex" rather than the "courseId".

The 2nd major section is a little more complex.
- It starts with 4 bytes representing the Grand Prix points scored in each CC mode
  - Each byte is broken up into dibits (2 bit groups). The least significant dibit is for the Mushroom cup, the next dibit is Flower cup, etc.
- Then there's 4 bytes of unknown purpose
- It then has 2 minor sub-parts, each comprised of:
  - The top 3-lap record for 8 courses
  - The top 1-lap record for 8 courses
  - 8 bytes of unknown purpose

Its unclear why its broken up into 2 parts each covering 2 cup's worth of tracks, instead of 4 parts 1-per-cup or even just 1 big part like the first major section.
This is finally followed by 8 more bytes of unknown purpose. They appear to be copies of the top 8 bytes at the front of the 2nd major section.

Each record is stored as 3 bytes in little-endian order (instead of big-endian like everything else).
After converting the record back to big-endian, it can be interpreted as:
- The top nibble is a character ID, indicating which character set that record
- The bottom 5 nibbles are the time of the record in question measured in centiseconds


include/common_structs.h
```diff
@@ -297,6 +297,56 @@ typedef struct {
...
+typedef struct {
+    // There are 4 cups total
+    CupTimeTrialRecords cupRecords[4];
+} AllCourseTimeTrialRecords; // size = 0x180
+
+typedef struct {
+    // Records 0 through 3 are for the first cup's courses
+    // Records 4 through 7 are for the second cup's courses
+    u8 best3laps[8][3];
+    u8 best1laps[8][3];
```
MegaMech, line 326: Consider:
bestSingleLaps, bestThreeLaps.

include/common_structs.h
```diff
@@ -297,6 +297,56 @@ typedef struct {
...
+    AllCourseTimeTrialRecords allCourseTimeTrialRecords;
+    // GP Points scored for each CC mode
+    // 1st place is 3 points, 2nd is 2, etc.
+    // Lowest dibit is the Mushroom Cup, 2nd dibit is the Flower Cup, etc
+    u8 grandPrixPoints[4];
+    // Its unknown what these bytes are used for
+    // They appear related to the Grand Prix points in some capacity but I can't figure out how
+    u8 unknownBytes1[4];
+    // For some reason there's 2 entries covering 4 cups
+    // Instead of 4 entires, one per cup. Or even just one big entry for all 4 cups
```
MegaMech, line 341: entries*

src/code_800AF9B0.c
```diff
@@ -3213,64 +3216,69 @@ GLOBAL_ASM("asm/non_matchings/code_800AF9B0/func_800B4DCC.s")
 
 #ifdef MIPS_TO_C
 //generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-s32 func_800B4DF4(void *arg0) {
-    return ((arg0->unk1 << 8) + arg0->unk0 + (arg0->unk2 << 0x10)) & 0xFFFFFF;
+// Convert time trial record from its little-endian form to big-endian
+s32 func_800B4DF4(u8 *arg0) {
+    return (arg0[0] + (arg0[1] << 8) + (arg0[2] << 0x10)) & 0xFFFFFF;
```
MegaMech, line 3221: In the future this could end up as a macro something like:
#define BIG_TO_LITTLE_ENDIAN ()

MegaMech (commented):
Wow, this one spun my brain around in circles.

There's a spelling mistake I noticed in a comment. The other comment is a suggestion.

Everything else lgtm.

PR #70, opened 2021-11-28
Text Color defines

Tmcg2:

### DIFF
--- a/asm/bss_8001C4D0.s
+++ b/asm/bss_8001C4D0.s
@@ -1892,7 +1892,7 @@ glabel D_8018E858
 glabel D_8018E85C
 .skip 4
 
-glabel D_8018E860
+glabel gTextColor
 .skip 8
 
 glabel D_8018E868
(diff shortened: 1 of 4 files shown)


MegaMech:
I'm wondering if BGR should be written slightly differently.
Everyone recognizes RGB. Perhaps most coders are used to swapping the letters around?

What about: TEXT_COLOUR_CYCLE
or
TEXT_CYCLE_COLOURS ?

Tmcg2:
I only chose BGR since super technically the cycle starts on Blue and then goes Green -> Red -> Blue, etc.
RGB is definitely the more common abbreviation and I'm not opposed to swapping it if desired.
I'm partial to having the colors of the cycle in the name since it doesn't cycle through all 4 colors, it (for some reason) leaves out yellow.

MegaMech:
I'm leaning towards:
TEXT_BLUE_GREEN_RED_CYCLE_1
It's longer but the benefit is that its meaning is immediately clear.

Which PR should be merged first or does it not matter?

Tmcg2:
Shouldn't matter, I did my best to make the independent of each other. I'll get to making the requested changes

PR #72, opened 2021-11-29
Fix bad merge conflict resolution in defines.h

Tmcg2:


Tmcg2:
I did do a ```make clean; make``` to ensure that this did fix the problem. Everything seems fine on my end

PR #73, opened 2021-12-01
Match a couple functions concerning string display width

Tmcg2:
Unfortunately get_character_width_index doesn't technically
match. Due to some linker alignment shenanigans its unlikely
it can be 100% matched until near the end of the project.

However, in its current state the only mismatches are in the
targets of its jump table. The actual code part matches


MegaMech:
My only thought on this one is perhaps:
`get_ascii_width_index` instead of
`get_character_width_index`
or some variation such as ascii_char

Mips to C switch statements often have repeated lines that don't belong btw. It will show up in all the cases.  Sometimes func calls have extra args that don't belong too.

PR #80, opened 2021-12-05
Match some functions related to printing text on screen

Tmcg2:
Some functions don't match 100% yet, but they're no longer in thier
 original m2c format either

Remove some extraneous "#if 1" directives leftover from an older
 commit


src/code_80091750.c
```diff
@@ -1079,361 +1078,214 @@ extern s8 gTextColor;
...
             }
-            phi_s2 = phi_s2_2;
-            if (*phi_s2_2 == 0) {
-                goto block_32;
+            else if ((characterWidthIndex != -2) && (characterWidthIndex == -1)) {
+                column += (tracking + 7) * x_scale;
             }
-            goto loop_22;
-        }
-        temp_v0_4 = gDisplayListHead;
```
CoderStig, line 1291: Unless this scenario is somehow different...
Whenever you see this pattern, it's a gfx macro that sends graphics commands straight to the gdp.

Take unk0 put it through `./gfxdis.f3dex` (found on github/n64-master) along with the argument. They must be 32 bits each: ex:
`06000000020077D8`

```
./gfxdis.f3dex -d 06000000020077D8
{
    gsSPDisplayList(0x020077D8),
}
```
Now we have to make this work in code:
Step 1: gsSP -> gSP (Do this for all commands when you're working in code or .text instead of .data. This is called dynamic display lists verses static display lists like what you see in the trophy file)
Next gDisplayListHead is always added as an argument with ++
    gSPDisplayList(gDisplayListHead++, 0x020077D8);

Sometimes there's an actual function hidden between a bunch of gfx. It's easy to spot, but if you're not looking for it it's easily missed.

Tmcg2: I'll give it a shot, the display list stuff is the major contributor to the non-matchiness of those functions so I'm all for fixing it. Thank you for the info!

Tmcg2: Something that may be useful for future reference
You can make gfxdis.f3dex output in the desired format automatically by using the -g option
```
gfxdis.f3dex -g gDisplayListHead++ -d 06000000020077A8 06000000020077D8
gSPDisplayList(gDisplayListHead++, 0x020077A8);
gSPDisplayList(gDisplayListHead++, 0x020077D8);
```

CoderStig: Waaaaaat?!? Seriously? *facepalm* I've been doing it manually this whole time.

Also,  since I missed this step: If `0x020077A8` is a variable/symbol then you can put `D_020077A8`. It's just you can't put the D_ part into gfxdis. You could even ignore typing in the symbol into gfxdis if you just want to remember to add it as the second argument. You'd still have to put eight zeros though.

Tmcg2:
I don't want to leave print_text1 in an unmatched state, but I really don't know how to get it matching.

PR #90, opened 2021-12-12
Identified some variables related to sound mode

Tmcg2:
Updated the SaveData structure as a couple unknown bytes were identified
Also updated some of the comments concerning the unknown bytes


include/common_structs.h
```diff
@@ -337,17 +337,21 @@ typedef struct {
     // 1st place is 3 points, 2nd is 2, etc.
     // Lowest dibit is the Mushroom Cup, 2nd dibit is the Flower Cup, etc
     u8 grandPrixPoints[4];
+    u8 soundMode;
     // Its unknown what these bytes are used for
-    // They appear related to the Grand Prix points in some capacity but I can't figure out how
-    u8 unknownBytes1[4];
+    u8 unknownBytes[3];
     // For some reason there's 2 entries covering 4 cups
```
MegaMech, line 343: Could these unk ones just be pointers back to the save data struct. Although, that would be pretty obvious because in-game debugger would show as variable address in that spot.

include/defines.h
```diff
@@ -95,4 +95,16 @@
...
+/**
+ * Sound mode options
+ * Option 2 appears to be unused, as such its probably not
+ * a valid option
+**/
+#define SOUND_STEREO      0
+#define SOUND_HEADPHONES  1
+#define SOUND_UNUSED      2
+#define SOUND_MONO        3
+#define NUM_SOUND_OPTIONS 4
```
MegaMech, line 108: You use soundMode but than  soundOptions here...

I think a better name for soundMode might exist. I'm sure there's a technical term for this kind of sound. Like soundSpeakerType or soundSetupOptions or something along those lines. This way people would immediately know what gSoundMode is.

include/common_structs.h
Tmcg2: I don't think so, they're purpose seems related to being a hash of sorts for the save data.
```func_800B492C``` seems to calculate something akin to a hash of grandPrixPoints and the screenMode. That value is stored in ```unknownBytes[1]``` by the looks of it.
```unknownBytes[2]``` seems to be somewhat similar, no idea what ```unknownBytes[0]``` could be for though.

include/defines.h
Tmcg2: I'm open to suggestions, names are not my strong suit.
soundSpeakerType is my favorite of what you suggested

MegaMech (approved):

PR #93, opened 2021-12-14
Rename a couple variables

Tmcg2:
Some to more accurately match the naming scheme used elsewhere
The other to more accurately identify its purpose


asm/bss_8001C4D0.s
```diff
@@ -2135,7 +2135,7 @@ glabel gCupSelection # 0x8018EE09
 glabel D_8018EE0A
 .skip 1
 
-glabel gCourseSelection # 0x8018EE0B
+glabel gCupTrackSelection # 0x8018EE0B
```
MegaMech, line 2138: We need to be consistent and choose one of Course or Track. I would recommend sticking with Course since the rest of the repo is named that way. `gCupCourseSelection`

If track is better than course I can be convinced. But I've been thinking `course` for so long I'm biased towards that.

asm/non_matchings/code_80091750/func_800A34A8.s
```diff
@@ -56,9 +56,9 @@ glabel func_800A34A8
 .L800A357C:
 /* 0A417C 800A357C 0C028E84 */  jal   func_800A3A10
 /* 0A4180 800A3580 27A40080 */   addiu $a0, $sp, 0x80
-/* 0A4184 800A3584 3C048019 */  lui   $a0, %hi(gGPOverallRanks) # $a0, 0x8019
+/* 0A4184 800A3584 3C048019 */  lui   $a0, %hi(gCharacterIdByGPOverallRank) # $a0, 0x8019
```
MegaMech, line 59: This one is very confusing. I don't have a better alternative though.

asm/bss_8001C4D0.s
Tmcg2: No, you're right. Since they're synonyms I swap between them all the time and forgot that ```course``` is the more canonical term in this repo.

asm/non_matchings/code_80091750/func_800A34A8.s
Tmcg2: I honestly don't like the name very much either, but I'm unable to find a more terse way of getting across what the array tracks (character IDs) and how it's indexed (overall rank in the current GP)

MegaMech (approved):

PR #105, opened 2021-12-18
Add gPlayerIsAccelerating, gPlayerBalloonCount and update related code

MaciekBaron:
Starting with a small pull request just to make sure I'm doing things correctly before pushing bigger changes.

The gist of this change is, there is a flag that changes to `1` when a player has acceleration pressed, `0` otherwise.

(Also removed stray white space)


MaciekBaron:
Realized `spawn_player` calls this and haven't updated that yet, marking as WIP

MaciekBaron:
Alright, I don't think I'm knowledgable enough to understand why I'm getting a mismatch.

When acceleration is pressed, only the 4th byte gets changed - it makes sense to me that the three previous bytes are used for something else, however I only get a match if that is treated as an `s32`.

MaciekBaron:
Changed it to s32 for now so that it matches

src/code_800393C0.c
```diff
@@ -17,7 +17,7 @@ extern Camera *camera4;
 
 //GLOBAL_ASM("asm/non_matchings/code_800393C0/spawn_player.s")
 
-void spawn_player(Player *player, s8 arg1, f32 arg2, f32 arg3, f32 arg4, f32 arg5, u16 arg6, s16 arg7) {
+void spawn_player(Player *player, s8 playerNumber, f32 arg2, f32 arg3, f32 arg4, f32 arg5, u16 arg6, s16 arg7) {
```
MegaMech, line 20: I would prefer `playerIndex`

include/common_structs.h
```diff
@@ -391,4 +391,8 @@ typedef struct {
     f32* unk_00[5];
 } struct_D_800E261C;
 
+typedef struct {
+    s32 playerRelatedFlags[8]; // Set to 1 when acceleration is pressed for each player
```
MegaMech, line 395: I'm not really a fan of this name and then having the comment being more specific. I'll take a closer look when I have some more time.

MegaMech (commented):
Hi MaciekBaron,

Thanks for your PR!
Just a heads up, that we must not have looked at n64 source code leaks. SDK content from CDs that were sold are generally considered okay though.

include/common_structs.h
MaciekBaron: I've done some more testing and I don't think those other bytes get ever written to, and they always stay 0. I think the game literally uses the whole 4 bytes for this one flag, which is weird, but there are weirder things in the code. I changed it from a struct to an s32.

MaciekBaron:
> Hi MaciekBaron,
> 
> Thanks for your PR! Just a heads up, that we must not have looked at n64 source code leaks. SDK content from CDs that were sold in auctions are generally considered okay though.

Thanks for the heads-up. I'm just using debugging tools to write the code, definitely not planning to do anything illegal.

I also now added some balloon related code to the same PR.

MegaMech:
This is definitely an s32 in this layout:

```
p1IsAccel, p2IsAccel, p3IsAccel, p4IsAccel
AIp5IsAccel, AIp6IsAccel, AIp7IsAccel, AIp8IsAccel
```

src/code_80057C60.c
```diff
@@ -11100,32 +11100,32 @@ GLOBAL_ASM("asm/non_matchings/code_80057C60/func_8006B87C.s")
...
+    s8 playerIndexTemp;
     u16 *temp_v0;
     u16 temp_t5;
 
-    temp_v1 = (arg1 * 2) + &D_8018D8C0;
-    temp_a3 = *temp_v1;
-    temp_a2 = arg1;
-    if (temp_a3 >= 0) {
-        temp_v0 = (arg1 * 6) + (temp_a3 * 2) + &D_8018D5F0;
+    currentPlayerBalloonCount = (playerIndex * 2) + &gPlayerBalloonCount;
```
MegaMech, line 11114: Is this  playerIndex[1]? or playerIndex++ ? or something like that.

src/code_80057C60.c
```diff
@@ -11100,32 +11100,32 @@ GLOBAL_ASM("asm/non_matchings/code_80057C60/func_8006B87C.s")
...
+    playerBalloonCountValue = *currentPlayerBalloonCount;
+    playerIndexTemp = playerIndex;
+    if (playerBalloonCountValue >= 0) {
+        temp_v0 = (playerIndex * 6) + (playerBalloonCountValue * 2) + &D_8018D5F0;
```
MegaMech, line 11118: playerIndex[5]?

MegaMech (approved):

MegaMech:
I was about to merge this but just realized a potential issue with the name.
Where as `gPlayerIsAccelerating` seems pretty accurate. The issue is when you're at max speed you're not really accelerating anymore. However, it equals 1 as long as the throttle button is held down. 


Try something like:
```
gPlayerIsThrottleActive
```

MaciekBaron:
Thanks, that's a good shout!

MegaMech (approved):

PR #107, opened 2021-12-20
Change some "char" terminology to "glyph"

Tmcg2:
This matches more closely with similar functions in SM64

Identified a LUT for the segmented address of each glyph's texture

CharWidthMap is more accurately called GlyphDisplayWidthMap. The values in
 the map don't exactly match the texture width for the given glyph. They
 likely have more to do with how the glyph is drawn on screen rather than
 being a measure of its true width.

There are other LUT-like entries in data_0DD0A0_2 that clearly contain
 glyph textures, but it's not clear how they differ from the "main"
 LUT.


asm/non_matchings/code_80091750/print_text1.s
```diff
@@ -121,13 +121,13 @@ glabel print_text1
...
 .L80093540:
-/* 094140 80093540 0C024B24 */  jal   get_ascii_char_width_index
+/* 094140 80093540 0C024B24 */  jal   char_to_glyph_index
 /* 094144 80093544 02402025 */   move  $a0, $s2
 /* 094148 80093548 04400027 */  bltz  $v0, .L800935E8
 /* 09414C 8009354C 00408825 */   move  $s1, $v0
 /* 094150 80093550 8FAC0074 */  lw    $t4, 0x74($sp)
-/* 094154 80093554 3C0B800E */  lui   $t3, %hi(D_800E7E84) # $t3, 0x800e
-/* 094158 80093558 256B7E84 */  addiu $t3, %lo(D_800E7E84) # addiu $t3, $t3, 0x7e84
+/* 094154 80093554 3C0B800E */  lui   $t3, %hi(gGlyphTextureLUT) # $t3, 0x800e
```
MegaMech, line 129: gLookupGlyphTexture

MegaMech (changes requested):

asm/non_matchings/code_80091750/print_text1.s
MegaMech: or 
gGlythTextureTable

I might actually prefer that one. If you ever think your reasoning is better though, definitely make the argument.

Tmcg2: The choice for LUT instead of just Table was mostly due to stuff like ```main_hud_lut``` in SM64. I'm fine with changing from LUT to Table, but I can see value in mimicking SM64 in this regard.

MegaMech (approved):

PR #109, opened 2021-12-20
Create Position struct, update player pointer labels and label some functions

MaciekBaron:
The way `func_802B51E8` is called shows that the player's position is treated like a struct. I therefore renamed the unknown struct to `Position` (it could be called Vector3 or something, but there's already a similar type, so called it Position for now). I then updated the player struct to use that new struct, and moved the struct to `common_structs.h` to avoid a circular include (happy to use a different approach).

Next, I relabelled the pointer that we know point to individual players - this makes the code a lot more understandable in a lot of areas. This allowed me to identify the purpose of at least one function.

Finally - apologies - but my editor cleans up stray white space in the files I edit. If you want, I can disable it for this project, as it might make reviewing difficult.


MegaMech:
I'm surprised this matches. I thought some recent code I decomped required array access... I guess that had nothing to do with the mismatch.
Despite the fact this matches. Given that sm64 and other n64 games have ran into issues with position struct access I highly recommend this:
```
Vec3f pos;
```
I know I didn't set it up like that, I just didn't get around to it. See the below sm64 MarioState struct for reference. When it makes sense, we're trying to copy sm64 so that people can transition between the two games. The Mario decompers strongly believe this is the correct. Given that mk64 appears to have copied sm64 it's also likely that this is the same.

https://github.com/n64decomp/sm64/blob/1372ae1bb7cbedc03df366393188f4f05dcfc422/include/types.h#L274

However, if you feel you have a stronger argument for  your way, feel free to convince me.

MaciekBaron:
Hmm, I see. I didn't know the background to this. I don't have a stronger argument, probably best to stick to it. I'll change that.

Also fyi I just reverted the trailing spaces changes on `nop` to make reading easier

MegaMech:
> Hmm, I see. I didn't know the background to this. I don't have a stronger argument, probably best to stick to it. I'll change that.
> 
> Also fyi I just reverted the trailing spaces changes on `nop` to make reading easier

Thanks! I don't have an issue with your cleaning up white spaces. An exception for nops would be nice though (Or a PR that does it to all the ASM, so I only have to look at that once).

MegaMech:
Also,

if you ever have any questions feel free to leave it in the  mk64-decomp channel on the Discord.

MaciekBaron:
I might need to look at this after Christmas, will change everything then

MaciekBaron:
Closing this for now, will just create a new PR as it will make things simpler

PR #113, opened 2021-12-22
Menu selection global variable

ChiefFruitcake:
-Rename D_800E86A0 to gMenuSelection
-Add menu selection options to defines.h


src/code_80091750.c
```diff
@@ -1889,20 +1889,20 @@ void func_80094A64(s32 arg0) {
     func_80092290(5, &D_8018E854, &D_8018E85C);
     func_80099A70();
     func_8009C918();
-    temp_t1 = D_800E86A0;
+    temp_t1 = gMenuSelection;
```
ChiefFruitcake, line 1892: First time making a PR for this project... I wasn't sure if I should be simplifying logic where appropriate (like removing the extra temp_t1 variable here)

ChiefFruitcake: I'm guessing I probably shouldn't modify any logic of the generated C code unless I get the entire function matching the assembly?

include/defines.h
```diff
@@ -70,6 +70,25 @@
 #define PEACH  6
 #define BOWSER 7
 
+/**
+ * Options for gMenuSelection
+**/
+#define START_MENU 0
+#define GAME_SELECT_MENU 1
+#define PLAYER_SELECT_MENU 2
+#define MAP_SELECT_MENU 3
```
MegaMech, line 79: MAP->COURSE

src/code_80091750.c
MegaMech: You're right that in-this case the temp_t1 doesn't exist. I wouldn't change these unless you were working on that function trying to match it. In-progress matches are fine. (So feel free to change things). But if you're using `./diff func_name`  you'll know if  something worked better or not.

MegaMech (changes requested):

include/defines.h
```diff
@@ -70,6 +70,25 @@
 #define PEACH  6
 #define BOWSER 7
 
+/**
+ * Options for gMenuSelection
+**/
+#define START_MENU_DUPLICATE 0
+#define GAME_SELECT_MENU_DUPLICATE 1
+#define PLAYER_SELECT_MENU_DUPLICATE 2
+#define COURSE_SELECT_MENU_DUPLICATE 3
```
ChiefFruitcake, line 79: I ended up swapping these lower values to the 'duplicates' since the values below are the ones that are actually being set in the code.

src/code_802A3730.c
```diff
@@ -150,18 +150,18 @@ void func_802A38B4(void) {
 
     switch(D_80150124) {
         case 0:
```
ChiefFruitcake, line 152: @MegaMech I think you had the right idea about quit out values for the 'duplicated' menu options. I haven't thoroughly tested but D_80150124 seems to be getting set to the duplicate value when quitting to course change or driver change from game.

MegaMech: Now that we know that. I think we should put that in the name. Something along the lines of:

START_MENU_FROM_QUIT

ChiefFruitcake: And it gets set to what I called UNUSED_MENU (4) when retrying from time trials... think I'll have to investigate some more...

src/code_8028DF00.c
```diff
@@ -234,16 +234,16 @@ void func_8028E3A0(void) {
     if (D_80150120) {
 
         if (gCupCourseSelection == CUP_COURSE_FOUR) {
-            D_80150124 = 5;
+            gMenuSelectionFromQuit = OPTIONS_MENU;
```
ChiefFruitcake, line 237: I was confused by the OPTIONS_MENU... it's actually brining the player to the winning ceremony cutscene.

ChiefFruitcake: D_800DC524 is getting set to gMenuSelectionFromQuit, then in thread5_game_logic's main loop, D_800DC50C is set to D_800DC524 if they aren't equal. func_80002684 has a switch statement on the value of D_800DC524 and case 5 (OPTIONS_MENU) calls load_ending_sequence_royalraceway().

include/defines.h
```diff
@@ -70,6 +70,27 @@
...
+/**
+ * Options for gMenuSelection and gMenuSelectionFromQuit
+**/
+#define START_MENU_FROM_QUIT 0
+#define GAME_SELECT_MENU_FROM_QUIT 1
+#define PLAYER_SELECT_MENU_FROM_QUIT 2
+#define COURSE_SELECT_MENU_FROM_QUIT 3
+#define RACING 4
+#define OPTIONS_MENU 5
+#define ENDING_SEQUENCE 5
```
ChiefFruitcake, line 82: Maybe move ENDING_SEQUENCE and CREDITS_SEQUENCE out of this list of defines into a separate list

MegaMech: Whatever seems to make the most sense I'm good with.

ChiefFruitcake: Think I'll leave it as is unless I can make some more sense out of the different variables using these values

MegaMech (approved):

PR #116, opened 2021-12-26
Texture related structs and some controller pak menu stuff

Tmcg2:
This branch started off as an attempt to identify some stuff related to related to the controller pak menu (reach by holding the start button on game boot)

It eventually morphed into defining a bunch of structs related to textures and identifying some other structs and arrays that hold those structs.

I can split this up into smaller PR's if desired, and as always I'm open to suggestions for names


asm/bss_8001C4D0.s
```diff
@@ -2082,7 +2082,7 @@ glabel D_8018EDEE
 glabel gDebugMenuSelection
 .skip 1
 
-glabel gPlayerCountSelection2 # 0x8018EDF0
+glabel gMenuState # 0x8018EDF0
```
Tmcg2, line 2085: I don't really know what to name this but its not related to player count at all. It only coincidentally matches the human player count on the main menu, its not used for that purpose though

ChiefFruitcake: It looks like its value corresponds to what is selected in the Controller Pak Menu and func_800B1C90 is handling what to do for controller inputs depending on the selection.

Tmcg2: That's part of it, its also has values that only show up when you delete an entry from the controller pack. It also appears to be used in the main menu, indicating which player-count mode you've chosen, which is probably why it was original named "PlayerCountSelection" because it does sort of look like that's what its tracking

MegaMech: Is the symbol used for two purposes? Or is the player-count theory wrong?
If this is for controllerPak. Would it make sense for everything to be named something like:
`gControllerPakState`?

asm/non_matchings/code_80091750/func_80091D74.s
```diff
@@ -77,8 +77,8 @@ glabel func_80091D74
 /* 092A84 80091E84 26310004 */  addiu $s1, $s1, 4
 /* 092A88 80091E88 1613FFF8 */  bne   $s0, $s3, .L80091E6C
 /* 092A8C 80091E8C AE22FFFC */   sw    $v0, -4($s1)
-/* 092A90 80091E90 3C108019 */  lui   $s0, %hi(D_8018EB80) # $s0, 0x8019
-/* 092A94 80091E94 2610EB80 */  addiu $s0, %lo(D_8018EB80) # addiu $s0, $s0, -0x1480
+/* 092A90 80091E90 3C108019 */  lui   $s0, %hi(gControllerPakPagesFree) # $s0, 0x8019
```
MegaMech, line 80: `gControllerPakNumPagesFree`?

asm/non_matchings/code_80091750/func_8009A6D4.s
```diff
@@ -0,0 +1,41 @@
+glabel func_8009A6D4
```
MegaMech, line 1: Add `# UNUSED` to the end of the glabel please.

asm/bss_8001C4D0.s
Tmcg2: It appears to be used in places other than the controller pak menu. Its just that it probably got its original name because if all you look at in the main menu it does look like its tracking player count. That's not what its doing, but I can see why it would be identified that way.

src/code_80091750.c
```diff
@@ -8682,25 +8612,16 @@ void func_8009F5E0(void *arg0) {
...
-                gDisplayListHead = temp_v0_2;
-                gDisplayListHead = temp_v0_2 + 8;
-                temp_v0_2->unk4 = 0x3200FF;
-                temp_v0_2->unk0 = 0xFA000000;
-                temp_v0_3 = func_8009B9D0(gDisplayListHead, &D_02001630);
-                gDisplayListHead = temp_v0_3;
-                gDisplayListHead = temp_v0_3 + 8;
-                temp_v0_3->unk4 = 0x323200FF;
-                temp_v0_3->unk0 = 0xFA000000;
+            // Setting the text color of the text on the controller pak menu
```
MegaMech, line 8615: Sets & Controller Pak

MegaMech (approved):

asm/bss_8001C4D0.s
ChiefFruitcake: I see D_8018EDF3 matching the player count on the main menu, but D_8018EDF0 is only a byte

PR #119, opened 2021-12-27
Naming global variables related to menu navigation

ChiefFruitcake:
I was investigating the usage of gMenuState which led to me renaming some global variables related to navigating the game menu and controller pak menu.


include/variables.h
```diff
@@ -25,7 +25,7 @@ extern s8    gHumanPlayerCount;
 extern f32   gKartFrictionTable[];
 extern f32   gKartBoundingBoxTable[];
 extern f32   gKartGravityTable[];
-extern s8    gMenuState;                            // D_8018EDF0
+extern s8    gControllerPakMenuSelection;           // D_8018EDF0
```
Tmcg2, line 28: I see why I thought this was related to the main menu now. I forgot this was an s8, not an s32, which led to me misinterpreting the debugger output. The name given here is much better

asm/bss_8001C4D0.s
```diff
@@ -2073,7 +2073,7 @@ glabel D_8018EDEB
 glabel D_8018EDEC
 .skip 1
 
-glabel D_8018EDED
+glabel gGameMenuSelectionDepth
```
MegaMech, line 2076: Depth is like: 1p,2p,3p,4p (depth of 1). VS, GP, Battle (depth of 2)? Yea, I think that's a good name.

Is `game` necessary? Or how about `MainMenu`?

asm/non_matchings/audio/external/func_800C2274.s
```diff
@@ -21,8 +21,8 @@ glabel func_800C2274
 /* 0C2EC0 800C22C0 AFA40048 */  sw    $a0, 0x48($sp)
 /* 0C2EC4 800C22C4 13200062 */  beqz  $t9, .L800C2450
 /* 0C2EC8 800C22C8 2406000F */   li    $a2, 15
-/* 0C2ECC 800C22CC 3C02800E */  lui   $v0, %hi(D_800DC530) # $v0, 0x800e
-/* 0C2ED0 800C22D0 8C42C530 */  lw    $v0, %lo(D_800DC530)($v0)
+/* 0C2ECC 800C22CC 3C02800E */  lui   $v0, %hi(gScreenModeSelection) # $v0, 0x800e
```
MegaMech, line 24: What does screenMode mean?

src/code_80057C60.c
```diff
@@ -742,7 +742,7 @@ void func_800591B4(void) {
                         }
                         func_8004F3E4(0);
                     }
-                    if ((D_800DC530 == 1) && (D_80165801 != 0)) {
+                    if ((gScreenModeSelection == SCREEN_MODE_2P_UD) && (D_80165801 != 0)) {
```
MegaMech, line 745: Avoid abbreviations such as UD. Using full/long names makes it so people always know what the meaning is. What's your philosophy behind UD/LR. Vs perhaps another way of writing it?

For instance why not `2PlayerMode`?

And just for clarification. My goal here is just to help us have the best names possible. so that someone not fluent in mk64 code can understand it. I'm not criticizing your approach, just helps me understand.

src/code_800AF9B0.c
```diff
@@ -2132,11 +2132,11 @@ void func_800B29D8(void *arg0, u16 arg1) {
                     sp1C = phi_v1;
                     func_800B44AC(temp_a0_2);
                     play_sound2(0x49008000);
-                    phi_t0_2 = &D_8018EDED;
+                    phi_t0_2 = &gGameMenuSelectionDepth;
                 }
             }
             if ((phi_v1 & 0x4000) != 0) {
-                *phi_t0_2 = 3;
+                *phi_t0_2 = PLAYER_AMOUNT_SELECTION;
```
MegaMech, line 2139: `ACTIVE_PLAYERS_SELECTION`?
`NUM_PLAYERS`? `PLAYERS_SELECTED`?

I would replace amount with `number`.

src/code_800AF9B0.c
```diff
@@ -2150,21 +2150,21 @@ void func_800B29D8(void *arg0, u16 arg1) {
                         if (temp_v1_3 != 1) {
                             if (temp_v1_3 != 2) {
                                 if (temp_v1_3 != 3) {
-                                    *phi_t0_2 = 6;
+                                    *phi_t0_2 = CONFIRMATION_SELECTION;
```
MegaMech, line 2153: Potentially `CONFIRMED_OK_SELECTION` as that would relate to what you see in-game. But it's just an idea. Feel free to ignore this comment.

asm/non_matchings/audio/external/func_800C2274.s
ChiefFruitcake: It controls how the screen is displayed during gameplay and the demos. So whether the screen is split for 2 players, 3+ players, or just 1 player.... also where the pause menu text appears.

src/code_80057C60.c
ChiefFruitcake: No worries, happy to get feedback since I would like this to be understandable. UD/LR is from the text on the debug menu, but I think you're right that it's worth spelling out up/down and left/right. Same for "Player" instead of "P"

src/code_800AF9B0.c
ChiefFruitcake: I think `NUM_PLAYERS` and `PLAYERS_SELECTED` may confuse this with the actual chosen number of players. This is meant to signify the player's location in the main menu is at the point where they are choosing 1P, 2P, 3P, or 4P. 

`PLAYER_NUMBER_SELECTION`?

src/code_800AF9B0.c
ChiefFruitcake: Yeah, I like that idea

asm/bss_8001C4D0.s
ChiefFruitcake: Yup, that's right.

`Game` matches the define I gave for this menu (`GAME_SELECT_MENU`), which was based on the text banner in the game saying 'Game Select'. Though I do think `MainMenu` is probably more intuitive. I can change this and the relevant #defines to `MainMenu`.

include/variables.h
ChiefFruitcake: Ahhh ok. Wasn't sure if I was missing something. Glad this isn't some weird dual-use variable

src/code_80057C60.c
MegaMech: I suppose this is worth having a debate over... I kinda think `SPLITSCREEN_HORIZONTAL` `SPLITSCREEN_VERTICAL`. Would maaaaybe be better. But there is an argument to be had, from a decomp perspective to use the in-game names.

src/code_800AF9B0.c
MegaMech: Yea, player num selection is fine.

src/code_80057C60.c
MegaMech: I'm fine with P1, P2, P3, P4 for flags. I'm pretty sure everyone will understand it. But if you're going to do that. Just make sure you use that in-places that make sense.

ChiefFruitcake: I like adding `SPLITSCREEN`, I think it makes it a bit clearer what `SCREEN_MODE` is actually about. 

So `SCREEN_MODE_1P`, `SCREEN_MODE_2P_SPLITSCREEN_HORIZONTAL`, `SCREEN_MODE_2P_SPLITSCREEN_VERTICAL`, `SCREEN_MODE_3P_4P_SPLITSCREEN`?

MegaMech: Yea, I think those work great.

MegaMech (dismissed):

MegaMech:
Resolve please and then I'll merge

PR #121, opened 2021-12-28
First decompile attempt. code_800AF9B0 

ChiefFruitcake:
Decompiled a few of the simpler non-matching functions in code_800AF9B0. Mostly to make sure I understand the process before I attempt to do anything larger.


src/code_800AF9B0.c
```diff
@@ -4,10 +4,26 @@
...
+void func_800B559C(s32 arg0);
+void func_800B5948(void);
+u32  func_802A7B70(u16 x);
+
+extern u8 D_8018ED10[4]; // Direct reference to the grandPrixPoints section of save data
+extern u8 D_8018ED11;
+extern u8 D_8018ED12;
+extern u8 D_8018ED13;
+extern s8 D_8018ED16;
+extern s8 D_8018ED17;
```
ChiefFruitcake, line 22: I see there's an include/variables.h file... Should these declarations be moved into there and this file includes it?

Tmcg2: I tend to only put named variables in that file, but I don't think there's any actual restrictions on that. If you find it useful to put it in there I'd vote for just doing it

MegaMech: The backstory here is. Somebody decomped a single file (for which I very much appreciate). And they did this for it.
I haven't paid much attention to it. If you find it useful, do whatever you want with it.

ChiefFruitcake: ok, I'll leave these here then

asm/bss_8001C4D0.s
```diff
@@ -2121,7 +2121,7 @@ glabel D_8018EDFB
 glabel D_8018EDFC
 .skip 4
 
-glabel gStartMenuCounterForDemos
+glabel gMenuTimingCounter
```
ChiefFruitcake, line 2124: I realized this is also used in other menus. For certain menus, once this reaches a certain value or multiple of a value, it will play an audio clip prompting the player to make a selection, hit ok, etc.

MegaMech (approved):

MegaMech:
Please resolve the conflicts and then I can merge.

MegaMech (approved):

PR #129, opened 2022-01-03
Match a variety of functions, identify a struct type and some related variables, regenerated some mips_to_c code

Tmcg2:
The names of some of the variables, namely ```gD_8018E118TotalSize``` and ```gNumD_8018E118Entries```, are super bad but I was unable to some up with anything better. I wholeheartedly welcome any suggestions


asm/bss_8001C4D0.s
```diff
@@ -1785,7 +1785,7 @@ glabel D_8018E060
 glabel D_8018E0E8
 .skip 40
 
-glabel D_8018E110
+glabel gD_8018E118TotalSize
```
MegaMech, line 1788: I sort of prefer to leave a comment `glabel gTotalSize # D_8018E118 `
But it's not a big deal.

Also, when you compile it creates a .map file. And you can find the location of symbols in there (as long as you have OK)

MegaMech (approved):

PR #130, opened 2022-01-03
Match several functions, add item values

MaciekBaron:
Matching `func_80022A98`, `func_800219BC`, `func_8001F980` (was already matching), `func_80295BF8`, `func_8006E8C4` and `func_80021C78`.

Also added item values.


src/code_8001F980.c
```diff
@@ -1209,40 +1207,23 @@ void func_8002276C(void) {
...
-    if ((arg0->unk0 & 0x8000) == 0x8000) {
-        func_80026A48(arg1);
-        func_800235AC(arg0, arg1);
-        temp_v0 = arg0->unkBC;
-        temp_v1 = temp_v0 & 0x4000000;
-        if ((temp_v1 == 0x4000000) || ((temp_v0 & 0x8000000) == 0x8000000)) {
-            if (temp_v1 == 0x4000000) {
-                func_80022B50(arg0, arg1, 0x8000000);
+void func_80022A98(Player* player, s8 arg1) {
+    if ((player->unk_000 & PLAYER_EXISTS) == PLAYER_EXISTS) {
```
MegaMech, line 1211: Does `&= PLAYER_EXISTS) {` match?

MegaMech (approved):

src/code_8001F980.c
MegaMech: If it does, and you feel like changing it feel free to change that. If you don't feel like it. Then I'll merge. lgtm.

MaciekBaron: I tried that before, unfortunately it doesn't seem to match.

MaciekBaron:
In reply to both comments, I originally used [this](https://strategywiki.org/wiki/Mario_Kart_64/Items) for reference, but how about we use the [official manual](https://m1.nintendo.net/docvc/NUS/USA/NKTE/NKTE_E.pdf) to standarise the names? So `SUPER_STAR` and `SUPER_MUSHROOM`?

MegaMech:
I think this one is worth opening to discussion in Discord.

MaciekBaron:
I've updated the item names now based on our discussion on Discord

MegaMech (approved):

PR #131, opened 2022-01-03
More matching in code_800AF9B0

ChiefFruitcake:
Additional work done in code_800AF9B0... matched more functions and named some variables related to selecting a game from the main menu.

Named functions related to loading and validating save data.


asm/non_matchings/code_80091750/func_8009F5E0.s
```diff
@@ -597,26 +597,26 @@ glabel L8009FD90
 glabel L8009FE9C
 /* 0A0A9C 8009FE9C 3C028019 */  lui   $v0, %hi(D_8018EDF3) # $v0, 0x8019
 /* 0A0AA0 8009FEA0 8042EDF3 */  lb    $v0, %lo(D_8018EDF3)($v0)
-/* 0A0AA4 8009FEA4 3C03800F */  lui   $v1, %hi(D_800E86AB) # 0x800f
-/* 0A0AA8 8009FEA8 3C06800F */  lui   $a2, %hi(D_800E86AD) # 0x800f
+/* 0A0AA4 8009FEA4 3C03800F */  lui   $v1, %hi(gGameTypeRowSelectionForNumPlayers) # 0x800f
+/* 0A0AA8 8009FEA8 3C06800F */  lui   $a2, %hi(gGameTypeSubMenuRowSelectionForNumPlayers) # 0x800f
 /* 0A0AAC 8009FEAC 00621821 */  addu  $v1, $v1, $v0
-/* 0A0AB0 8009FEB0 806386AB */  lb    $v1, %lo(D_800E86AB)($v1) # -0x7955($v1)
+/* 0A0AB0 8009FEB0 806386AB */  lb    $v1, %lo(gGameTypeRowSelectionForNumPlayers)($v1) # -0x7955($v1)
```
ChiefFruitcake, line 603: Open to suggestions for these names.

gGameTypeRowSelectionForNumPlayers is an array storing which row is selected for each of 1P Game, 2P Game, etc. The value stored is an index, not the actual value of the game type. (e.g. gGameTypeRowSelectionForNumPlayers[3] == 0 means versus is selected in the 3P game menu).

gGameModeFromNumPlayersAndGameTypeRowSelection is a 2D array storing what game type each row of the 1P Game, 2P Game, ... menus actually corresponds to. So gGameModeFromNumPlayersAndGameTypeRowSelection[3][0] == 2 (Versus).

gGameTypeSubMenuRowSelectionForNumPlayers is a 2D array storing what row is chosen in the submenu for each game type in each 1P Game, 2P Game, ... menu.
E.g. gGameTypeSubMenuRowSelectionForNumPlayers[1][0] == 2 means 150CC mode in 1P Game, Grand Prix is selected. gGameTypeSubMenuRowSelectionForNumPlayers[2][1] == 0 means 50 CC mode in 2P Game, Versus is selected.

src/code_800AF9B0.c
```diff
@@ -3401,12 +3303,9 @@ void func_800B5404(s32 arg0, s32 arg1) {
 GLOBAL_ASM("asm/non_matchings/code_800AF9B0/func_800B5404.s")
 #endif
 
-u8 func_800B54EC(s32, s32);
-u8 func_800B5508(s32, s32, s32);
-
 // Get Grand Prix points for a given cup and CC mode
 u8 func_800B54C0(s32 cup, s32 cc_mode) {
-    return func_800B54EC(cup, D_8018ED10[cc_mode]);
+    return func_800B54EC(cup, *(&D_8018ED10 + cc_mode));
```
Tmcg2, line 3308: D_8018ED10 is definitely a weird variable. Here its clearly meant to be used like its the head of an array, but almost everywhere else its just a single u8 for the 50 CC Grand Prix points. I don't know if there's a clean way to have it do both those things

ChiefFruitcake: Yeah, I went back and forth between leaving it as an array or a u8. I thought it was a bit nicer as a u8 for consistency since there's a few places where D_8018ED11, D_8018ED12, and D_8018ED13 are referenced directly, but I don't feel too strongly either way.

src/code_800AF9B0.c
```diff
@@ -3034,67 +2958,59 @@ void func_800B4A9C(s32 arg0) {
...
 }
-#else
-GLOBAL_ASM("asm/non_matchings/code_800AF9B0/func_800B4CB4.s")
-#endif
 
-#ifdef MIPS_TO_C
-//generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-void func_800B4DCC(void *arg0, u32 arg1, s32 arg2) {
-    u32 temp_t8;
+void write_time_trial_record(u8 *timeTrialRecord, u32 time, s32 characterId)
```
ChiefFruitcake, line 3001: Maybe populate_time_trial_record would be a better name? Avoid confusion since no eeprom writes are called in this function

src/code_800AF9B0.c
Tmcg2: I toyed around with the idea of leaving it as a u8 and then doing ```#define allGpPoints (&D_8018ED10)``` but that's pretty ugly

asm/non_matchings/code_80091750/func_8009F5E0.s
MegaMech: That's pretty reasonable. I'm sort of confused in how `GameType` differs from `GameMode`.

src/code_800AF9B0.c
MegaMech: Maybe it's a union?
There's a hack you can do to have struct and array access using a union. But I suppose that doesn't quite fit this scenario...

src/code_800AF9B0.c
MegaMech: I agree

include/defines.h
```diff
@@ -21,6 +21,12 @@
 // Compiles to -0x1000 in diff.py
 #define PLAYER_HUMAN_AND_CPU PLAYER_EXISTS | PLAYER_HUMAN | PLAYER_CPU | PLAYER_START_SEQUENCE
 
+// 0 players, 1 player, 2 players, 3 players, or 4 players
+#define NUM_PLAYER_NUM_SELECTIONS 5
```
MegaMech, line 25: NUM used twice here intentional?

asm/non_matchings/code_80091750/func_8009F5E0.s
ChiefFruitcake: Ah... they're the same. I'll use `GameMode` to keep things consistent

include/defines.h
ChiefFruitcake: Yeahhh I was thrilled with this name. It's meant to be the number of possible 'num players'

src/code_800AF9B0.c
ChiefFruitcake: I'll try a union and see what you guys think

src/code_800AF9B0.c
```diff
@@ -2987,7 +2995,7 @@ void validate_save_data(void) {
             gSaveDataSoundMode = gSaveDataSoundModeBackup;
             D_8018ED16 = compute_save_data_checksum_backup_1();
             D_8018ED17 = compute_save_data_checksum_backup_2();
-            osEepromLongWrite(&gSIEventMesgQueue, (((u32) &D_8018ED10 - (u32) &D_8018EB90) >> 3), &D_8018ED10, 8);
+            osEepromLongWrite(&gSIEventMesgQueue, (((u32) &D_8018ED10 - (u32) &D_8018EB90) >> 3), (u8*) &D_8018ED10, 8);
```
ChiefFruitcake, line 2998: So I added the union, but this got me thinking... 
This call is meant to write 8 bytes starting at D_8018ED10 (4 bytes for the grandPrixPoints, 1 for the soundMode, and 3 for the checkSum), but with the union D_8018ED10 is now 4 bytes on its own?

Or does these lines in asm/bss_8001C4D0.s mean that D_8018ED11 will still be one byte after D_8018ED10?
glabel D_8018ED10
.skip 1

src/code_800AF9B0.c
```diff
@@ -6,6 +6,12 @@
 #include "global.h"
 #include "common_structs.h"
 
+union GrandPrixPointsUnion
+{
+    u8 grandPrixPointsArray[4];
+    u8 grandPrixPointsMushroomCup;
+};
```
ChiefFruitcake, line 13: Here's the union. Let me know if you think there's a better place for it. D_8018ED10 is only referenced in this files so I put it here.

Also if we want to keep the union, I should probably make D_8018ED88 (save data grandPrixPointsBackup) the same type.

include/defines.h
MegaMech: `MAX_NUM_PLAYERS`?
`NUM_PLAYERS_MAX`?
`FIVE_NUM_PLAYERS_SELECTED`?

Because this is a define remember that it's naming should be based upon the `5` I think. Since it's not a variable. Generally, we should understand what the define is without necessarily going to its definition.

ChiefFruitcake: I think most people would assume `MAX_NUM_PLAYERS` and `NUM_PLAYERS_MAX` to be 4.

Maybe:
#define ZERO_PLAYERS_SELECTED 0
#define ONE_PLAYERS_SELECTED 1
#define TWO_PLAYERS_SELECTED 2
#define THREE_PLAYERS_SELECTED 3
#define FOUR_PLAYERS_SELECTED 4
#define NUM_PLAYER_SELECTIONS 5 ?

MegaMech: TOTAL_SELELECTED_PLAYER_DEFINES ?

ChiefFruitcake: Works for me

src/code_800AF9B0.c
ChiefFruitcake: Quick investigation looking at build/us/mk64.us.map and it appears making D_8018ED10 a union of size 4 is fine

include/defines.h
```diff
@@ -21,6 +21,16 @@
 // Compiles to -0x1000 in diff.py
 #define PLAYER_HUMAN_AND_CPU PLAYER_EXISTS | PLAYER_HUMAN | PLAYER_CPU | PLAYER_START_SEQUENCE
 
+#define ZERO_PLAYERS_SELECTED 0
+#define ONE_PLAYERS_SELECTED 1
+#define TWO_PLAYERS_SELECTED 2
+#define THREE_PLAYERS_SELECTED 3
+#define FOUR_PLAYERS_SELECTED 4
+#define SELELECTED_PLAYER_DEFINES_TOTAL 5
```
MegaMech, line 29: I believe you took my review comment slightly too literally: `SELELECTED`

ChiefFruitcake: Haha whooops... thanks for the catch

MegaMech (approved):

PR #132, opened 2022-01-04
Add a couple header files, add some mips_to_c of some previously un-decomped functions

Tmcg2:
These files contain function prototypes for matches functions, NON_MATCHING functions,
 and functions used by matched functions.

At this point in time those headers may contain prototypes for functions defined in
 unrelated files, but fixing that will take time


MegaMech:
Please put header files that are for a specific .C file beside the .C file.

`/include/` is for SDK libraries and standard C headers. Or things that are generic like variables.h, common_structs, and types.h. The aforementioned of which likely need to be sorted at some point later on in the project when we have a better idea of where things likely should go.

PR #133, opened 2022-01-08
code_800AF9B0 - ghost data

ChiefFruitcake:
Decompiled some functions in code_800AF9B0.c that seem to be related to ghost data?


include/common_structs.h
```diff
@@ -401,8 +401,14 @@ typedef struct {
 typedef struct
 { 
     // Something related to time trial ghost data?
-    char unk_00[0x80];
-} struct_8018EE10_entry;
+    /* 0x00 */ s32 unk_00;
+    /* 0x04 */ u8  unk_04;
```
Tmcg2, line 405: I remember looking at this at some point and I think I know what unk4, 5, and 6 are
unk4 seems to be an indicator of whether there's a ghost in that slot. There's 2 ghost slots in total and this byte seems to indicate that the slot is in-use
unk5 is the course index of the map the ghost is for. Course "index" is different from course "id". Course "index" is the "TIME_TRIAL_DATA_*" entry for that map
unk6 is the character id of the ghost

I think unk00 and unk7F are some for of checksums, but I'm less certain about that

ChiefFruitcake: Thanks for the info. I agree that unk7F looks like a checksum. I'll look into the others and see if I can name some of these fields

ChiefFruitcake: Pushed a commit with some updated names. I wasn't sure about unk00... 

In `func_800B6178`, it's getting set to `D_80162DFC`, which is set to `D_8018CA78` in `func_80005B18`. `D_8018CA78` appears to be a frame counter so maybe unk00 is the length of the ghost data in frames? Not sure...

MegaMech (approved):

PR #134, opened 2022-01-08
Update the GfxPool struct and match some functions

Tmcg2:
Also reordered variables.h a bit


src/main.h
```diff
@@ -2,13 +2,14 @@
...
-#define GFX_POOL_SIZE 20836
+
+#define MTX_POOL_SIZE 0x0683
+#define GFX_POOL_SIZE 0x1D4C
 struct GfxPool {
-    Gfx buffer[GFX_POOL_SIZE];
-    struct SPTask spTask;
-};
+    /* 0x00000 */ Mtx matrixPool[MTX_POOL_SIZE];
+    /* 0x1A0C0 */ Gfx gfxPool[GFX_POOL_SIZE];
```
MegaMech, line 10: SM64's struct is laid out as such:
```
struct GfxPool {
    Gfx buffer[GFX_POOL_SIZE];
    struct SPTask spTask;
};
```
Thoughts on `gfxPool` vs just `buffer`? Probably don't need to include the struct name in the var name.

Tmcg2: I felt it would help with readability if the member name had some sort of indicator of what type it is. The member being named the same as the struct type is a little weird, but I don't find it terribly confusing. I'm open to suggestions though

MegaMech: I think buffer better describes what it is. And when we see this being used you generally see: `gfxPool->buffer` anyway
ex: `&gGfxPool->buffer[32])` vs ` gGfxPool->gfxPool;`

MegaMech: Not sure what the name the mtx tho

MegaMech: Okay, what you have is fine. But rename ` matrixPool[MTX_POOL_SIZE];` to `mtxPool[]`

Tmcg2: lol, just reverted gfxPool back to buffer. I'll update mtxPool, should I leave gfxPool as buffer or should I undo the revert?

Tmcg2:
Think I'm done adding commits now. Won't add anything else unless addressing a comment requires it

MegaMech (approved):

PR #138, opened 2022-01-14
code_8008C1D0 part 2

ChiefFruitcake:
- Matched more functions in code_8008C1D0
- Tried to name a few of the fields in the Player struct
    - boostVelocity: (velocity might not be the right term? Seems to control how much impact a mushroom/well-timed start boost gives
    - boostTimer: Controls how long the player is under the effect of the boost
    - frontLeft, frontRight, backLeft, backRight TireTerrain: The terrain (maybe surface would be a better term?) that each tire of the player's kart is on (or above?)


include/common_structs.h
```diff
@@ -240,31 +239,31 @@ typedef struct {
     /* 0x0198 */ f32 unk_198;
     /* 0x019C */ f32 unk_19C;
     /* 0x01A0 */ f32 unk_1A0;
-    /* 0x01A4 */ s8 unk_1A4;
+    /* 0x01A4 */ s8 frontLeftTireTerrain;
```
MegaMech, line 242: I do lean towards `surface` because this term is used in Overkart 64. This allows the code-base to use similar terminology as standardized by the Mario Kart modders.

MegaMech: Also, it's spelled tyre, no?

ChiefFruitcake: Surface sounds good. And tire/tyre is American vs British spelling. I don't have any attachment to either though

include/common_structs.h
```diff
@@ -172,7 +171,7 @@ typedef struct {
     /* 0x0078 */ s16 unk_078;
     /* 0x007A */ s16 unk_07A;
     /* 0x007C */ s32 unk_07C;
-    /* 0x0080 */ f32 unk_080;
+    /* 0x0080 */ f32 boostVelocity;
```
ChiefFruitcake, line 174: Played around with this and I think a better term would be `boostPower` or `boostStrength`. When the player gets a boost, this field ramps up to a value and stays there until `boostTimer` reaches zero, and then it ramps down to 1. Once it reaches one the boost ends. It doesn't appear to be a direct measure of velocity

include/common_structs.h
MegaMech: Hmm... I think tyre is the global spelling as I'm not British and we spell it this way. I'm wondering if someone might confuse the symbol for deterioration.

include/common_structs.h
MegaMech: Ah, I remember exploring this one now. Yea, I'm fine with either boostPower or boostStrenth.

MegaMech (approved):

PR #145, opened 2022-01-18
Fix a small discrepancy with kart_attributes

Tmcg2:
Some of the variables defined in kart_attributes.c had they're type wrong when
 being used in other files.

This means that some previously defined variable no longer exist, however
 that's fine as they were never directly referenced. Their existence was
 almost certainly a decomp artifact.

For example, D_800E2404 was never "real". Instead it was likely a compiler
 artifact caused by accessing D_800E2400 at the hard coded index of 1.

I've updated kart_attributes.h to reflect these changes


MegaMech:
Nice catch!

MegaMech (approved):

PR #148, opened 2022-01-23
decomped some of 800431B0.c

MegaMech:
Also split an unused func out of 80040e50


src/code_80040E50.c
```diff
@@ -1097,6 +1097,9 @@ void func_80042E00(s32 arg0, s32 arg1) {
 #else
 GLOBAL_ASM("asm/non_matchings/code_80040E50/func_80042E00.s")
 #endif
+   
+// UNUSED
+GLOBAL_ASM("asm/non_matchings/code_80040E50/func_80042EB8.s")
```
ChiefFruitcake, line 1102: Should we be generating the mips_to_c for unused functions?

Tmcg2: I usually do, if only because some day far in the future even unused functions will have to be decomped in order to reach 100%

PR #152, opened 2022-01-27
Match OS funcs

MegaMech:
Matched osSyncPrintf, contram read / write, func_800CE720 (osPfsInit), and func_800CE7D4 (__osPfsGetStatus).

This leaves osLeoInterrupt as the last remaining libultra file.


src/os/osSyncPrintf.c
```diff
@@ -1,14 +1,14 @@
 #include "libultra_internal.h"
 #include "libc/stdarg.h"
 
-// neither function matches
-
-void osSyncPrintf(const char *fmt, ...) {
-    if (1);
+s32 osSyncPrintf(s32 arg0, s32 arg1, s32 arg2) {
```
Tmcg2, line 4: To be slightly nit-picky, _Printf requires the function pointer argument to return a `char *`, this should also return a `char*`
```s32 _Printf(char *(*prout)(char *, const char *, size_t), char *dst, const char *fmt, va_list args);```

I think we can deduce that arg0 is is a `char*`, arg1 is a `const char*`, and arg2 is a `size_t`

PR #157, opened 2022-02-03
Break out some credit text stuff into a C file

Tmcg2:


MegaMech (commented):
data_credits_128480 appears to be deleted but is still referenced in the loader file.

Perhaps rename 128480_1.s to 128480.s?

Does `credits_01` need the `01`?

Tmcg2:
The isn't deleted, its just has the bulk of the old data in it moved out. It went from having around 1000 lines to around 100

The `01` is there since I suspect `data_credits_128480_1.s` can be split up even further, so there could conceivably be `credits_02` and `credit_03`, etc in the future. I'm open to a different naming scheme if there's one you have in mind.

MegaMech:
image

MegaMech (approved):

PR #158, opened 2022-02-04
Make some asm files match the rest of project

Tmcg2:
Some asm files were in a slightly different format from the rest
 of the asm files in the project.


MegaMech:
These are handwritten asm. Their current formatting is intentional.

Tmcg2:
Oh I didn't realize. This PR can be declined if that's that case then. No problems are being fixed so its NBD if the changes aren't taken.

PR #161, opened 2022-02-12
Name functions that search for 8018D9E0 entries

Tmcg2:
The one with "_dupe" in its name is the one that has the potential hard lock
 laying around


MegaMech:
Is it worth putting @bug comment on that func and explain why the potential hardlock?

src/code_80091750.c
```diff
@@ -14780,6 +14780,17 @@ s32 *func_800AAEB4(s32 arg0) {
...
+ * function will enter a `while(1)` loop, hard-locking the
+ * game.
+ * 
+ * In practice this never appears to happen, but that is
+ * probably as much a matter of luck as it is good
+ * reasoning on the original author(s) part.
+**/
 struct_8018D9E0_entry *find_8018D9E0_entry_dupe(s32 arg0) {
     struct_8018D9E0_entry *entry = D_8018D9E0;
     for (; !(entry > (&D_8018D9E0[D_8018D9E0_SIZE])); entry++) {
```
MegaMech, line 14796: Sm64 has something similar for: `allocate_object` func.
```
// We've met with a terrible fate.
            while (TRUE) {
```

MegaMech: Now that I see the context. I guess it's not really a bug (but possibly). But the explanation could still be good.

This is probably good enough for now. Should I merge or do you have more changes?

Tmcg2: Nah, I don't plan to add anything else to this PR

PR #163, opened 2022-02-14
Match the other D_80189DE0 search functions

Tmcg2:


MegaMech:
Since we're back on this topic...

The while(1) is clearly intentional. Also, can we change `while(1)` to `while(TRUE)`?

As such, it's not really bug. If it went into the while loop unintentionally then I think bug would be fair. But at this time they didn't have assert or proper bug checking really if something was to result in crash. Or they were just lazy. So this is basically their method of stopping the game if something went wrong.

As such, I think it would be better to phrase the comment similar to the sm64 thing. Where it moreso acknowledges that if the object type is incorrect that the game `meets a terrible fate` or crashes or whatever.

Tmcg2:
Done

MegaMech (approved):

PR #168, opened 2022-02-20
Name arrays of character animations

Tmcg2:
Things like Single/Dobule blink animations for example


MegaMech:
The only thing confusing me with this PR is if this is end of the game celebration/lose animations Or if this is like player select menu animations (or potentially both). The reason I'm confused is cause of "deselect animation"

Tmcg2:
Celebration is used when you select a character, win a VS  or Battle round, or win a Grand Prix
Defeat is used when you lose a VS or Battle round (and, presumably, if you get 4th place in a Grand Prix)
Deselect is used when you deselect on the character on the character selection screen (so, press A to select then press B)

MegaMech (approved):

PR #170, opened 2022-04-12
Match func_80030FC8

Tmcg2:

D_80165460 appears to track on a per-player index whether they're doing regular or "fast" accleration
"fast" acceleration is triggered by triple tapping "A" instead of just pressing-and-holding
gKartAccelerationTables may actually be something more like gKartJerkTables, with player->unk_09C being player accleration
gKartTable800E36B0 is some sort of per-character multiplier to the jerk that is only used when triple-tap "A" is used
player->unk_214 is possibly the max-acceleration allowed
No speculation on what player->unk_0C4 could be


src/code_80027D00.c
```diff
@@ -4675,6 +4679,8 @@ GLOBAL_ASM("asm/non_matchings/code_80027D00/func_80030C34.s")
 
 #ifdef MIPS_TO_C
 //generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
+// This is likely the function responsible for detecting triple-tap "A"
+// for "fast" acceleration
```
MegaMech, line 4683: Is it intentional that you marked two functions with this comment? If so, it's fine. Just checking.

Tmcg2: Yes, although I should have made that clearer. It looks like there are 2 functions that each touch D_80165460 and while I haven't looked at them too carefully they both seem to do the same thing. So there might be slight differences, or maybe they're actually copies of each other.

Tmcg2:
Sorry for the extra commits, I should be done messing with this PR now

MegaMech (approved):

PR #175, opened 2022-04-15
Match dma_copy functions

Tmcg2:
I question the type of these ROM Segments. Based on their usage they look like
 they should be uintptr_t.

I also question the practice of &_texture_blah_blah[index] instead of plain
 _texture_blah_blah[index].

But, if it matches it matches


Tmcg2:
Actually, ignore that commit message, I see my mistake here. I got confused by the `&`. The segment rom array can be typed as u8 just fine because we're getting the address of a byte and passing that down to the OS as a uintptr_t. The segment rom array is not itself a list of uitnptr_t's.

MegaMech (approved):

PR #178, opened 2022-04-18
Mostly match a few functions in memory.c

Tmcg2:
They don't match quite right though, there's some incorrect
 regalloc occuring.


MegaMech:
I'm gonna have a closer look. But in my uneducated opinion I think the issue is this: &(*D_8015F580)[arg0];

Tmcg2:
There may be a better way to do it, but based on my understanding that's the only way to do it.
`D_8015F580` is a pointer to the head of an array.
Its not the head of an array itself, which is admittedly kind of weird. As some trivia, `D_8015F588` is the number of elements in the array.
The type of the array is `mk64_surface_map_ram`.
So, as best as I understand it the only way to declare `D_8015F580` is:
`mk64_surface_map_ram (*D_8015F580)[]`
Which means accessing it is super fugly.

MegaMech:
D_8015F580 isn't mk64_surface_map_ram tho. It's just a pointer.

It's probably an s32?

Tmcg2:
Seems you were right and I was overthinking it. While it would be more "accurate" to treat `D_8015F580` as a pointer-to-an-array, a plain pointer actually meets the regalloc requirements. Makes the code look nicer too.

MegaMech:
Nice!
Can we put these on the same line? Potentially re-name "thing"
`mk64_surface_map_ram *thing = &D_8015F580[arg0];`

However, if you don't have time, we can call it good as is.

MegaMech (approved):

PR #180, opened 2022-04-20
Match func_8002AB70

Tmcg2:
Added a header for another file, which required adjusting
 a couple other files


src/framebuffers.h
```diff
@@ -5,10 +5,10 @@
 #include "config.h"
 
 extern u8 D_802BA290[22768];
-extern u8 D_802BFB80[77824];
+//extern u8 D_802BFB80[77824];
```
MegaMech, line 8: Why is this commented out?

Tmcg2: Long story short it was to try and rectify a problem with including both variables.h and framebuffers.h in the same file. I've found a different way to address that problem in this PR, but I'm going to try and address the problem head on in a later PR.

MegaMech (approved):

PR #182, opened 2022-04-23
Fix some of the memory allocation in framebuffers

Tmcg2:
Also includes a the questionable change of undoing a previously
 made match for func_80027A20. The changes made here remove an
 AVOID_UB usage and (I would argue) is therefore more "correct",
 but undoing a byte-for-byte match is admittedly bad.

Also adds in some header files and updates some not-quite-right
 decomps to be slighly more accurate.


Tmcg2:
As mentioned in the commit message, this PR includes the highly questionable choice to undo a match made in the recent past. I'm not sure there's a way to rectify the rest of the changes I've made with the AVOID_UB of the current match. I've cooked my changes in the permuter for a couple hours by it didn't fine a way to resolve the regalloc differences.

I don't really know what to do about that :/

src/code_80027040.c
```diff
@@ -1,189 +1,10 @@
...
-/*
-u16 D_800DDEB0[] = { // sizes
-    0x06c0, 0x06e0, 0x06e0, 0x0680, 0x07c0, 0x0700, 0x0680, 0x0910
-};
-*/
-extern s32 gActiveScreenMode;
-
-extern uintptr_t **gKartTextureGroup18s;
-
-extern uintptr_t *gKartMarioGroup8[];
```
MegaMech, line 26: If this stuff should eventually be declared here (and not just extern'd) can this still be done in the .h file? or does it have to be in C?

src/code_80027040.c
```diff
@@ -290,39 +111,30 @@ uintptr_t ***gKartTextureTable1[] = {
...
-            osPiStartDma(&gDmaIoMesg, 0, 0, &_kart_texturesSegmentRomStart[SEGMENT_OFFSET(gKartTextureTable0[player->characterId][player->unk_24C[arg2]][player->unk_244[arg2]])], buf, D_800DDEB0[player->unk_0BC], &gDmaMesgQueue);
+            osInvalDCache(temp_s0, D_800DDEB0[player->characterId]);
+            osPiStartDma(&gDmaIoMesg, 0, 0, &_kart_texturesSegmentRomStart[SEGMENT_OFFSET(gKartTextureTable0[player->characterId][player->unk_24C[arg2]][player->unk_244[arg2]])], temp_s0, D_800DDEB0[player->characterId], &gDmaMesgQueue);
             osRecvMesg(&gDmaMesgQueue, &gMainReceivedMesg, 1);
         }
-    } else if (((((player->unk_0BC & 0x400) == 0x400) || ((player->unk_0BC & 0x1000000) == 0x1000000)) || ((player->unk_0BC & 0x2000000) == 0x2000000)) || ((player->unk_0BC & 0x10000) == 0x10000))
-    {
-        buf = &D_802DFB80[arg4][arg3][arg1 * 0x920];
-        osInvalDCache(buf, 0x780);
-        osPiStartDma(&gDmaIoMesg, 0, 0, &_kart_texturesSegmentRomStart[SEGMENT_OFFSET(gKartTextureGroup18s[player->characterId][player->unk_0A8 >> 8])], buf, 0x900, &gDmaMesgQueue);
```
MegaMech, line 320: Definitely a weird section. I was gonna have another look at it but you beat me to it lol.
(I actually mean the section like way below here where your comment is oops)

It's gonna be a few days before I finish reviewing this PR. Mainly because I want to try some things.

src/framebuffers.h
```diff
@@ -3,13 +3,25 @@
...
+    char unk_00[0x1000];
+} struct_D_802BFB80; // size = 0x1000
+
+typedef struct {
+    char unk_00[0x920];
+} struct_D_802DFB80; // size = 0x920
+
+typedef struct {
+    char unk_00[0x200];
+} struct_D_802F1F80; // size = 0x200
```
MegaMech, line 18: I'm fine with this change. Is the purpose of it to make it look cleaner?

src/code_80027040.c
```diff
@@ -331,51 +143,41 @@ GLOBAL_ASM("asm/non_matchings/code_80027040/func_80027040.s")
...
+#ifdef INCORRECT_REGALLOC
 void func_80027A20(Player *player, s8 arg1, s8 arg2, s8 arg3) {
-    // @bug: Undefined Behaviour UB
-    // untested fix
-    #ifdef AVOID_UB
-    void *temp_s0 = &D_802F1F80[arg3][arg2][0];
-    #else
-    void *temp_s0 = &D_802F1F80[arg3][arg2][arg1 << 8];
-    #endif
+    struct_D_802F1F80 *temp_s0 = &D_802F1F80[arg3][arg2][arg1];
```
MegaMech, line 176: Does this func match with `[arg1<< 8]`?

Tmcg2: Unfortunately no :(

src/framebuffers.h
Tmcg2: As best as I can tell its closer to being accurate to the actual underlying data. I doubt the underlying data is actually just a struct around a 0x200 sized char array, but its better than a plain char array.

If you look at `func_80027A201` in `code_80027040.c` you can see that `osPiStartDma` tries to copy 0x200 bytes from `_kart_texturesSegmentRomStart` to `D_802F1F80`. So, as best as I can tell, entries in `D_802F1F80` need to be 0x200 bytes large.

src/code_80027040.c
Tmcg2: I don't understand the question.
If the question is:
In the future can the currently non-initialized be put in this .c file or do they have to go elsewhere?
Then I think the answer is they (and their extern'd declarations) have to have their own file

MegaMech: Ah, ya, I see what's going on now. This change is good.

Eventually all of this data should be declared here. (Perhaps we could make a new file for it as long as it's linked just before its code file. This may or may not match like this). It's just sooo much data. I didn't do it all.

Tmcg2:
I'm still not sure what's going on with `D_802F1F80`, but I have found a good lead to follow.
In `func_800248D0` of 'code_8001F980` a single entry of `D_802F1F80` is saved to `D_80164B04`. `D_80164B04` is then used as part of a `gDPLoadTLUT_pal256` call. As best as I can tell right now, this means that `D_80164B04` (and by extension, the entries in `D_802F1F80`) are meant to be texture lookup tables (TLUT)

I don't know what to make of this yet, but I plan to investigate it further

MegaMech:
If you feel like it, you could take another look at this PR. I have matched all three DMA funcs now. Plus, I now remember what was going on in this file. So it's more likely I can look at your proposed changes and consider them, plus see if there's an alternative way to write the function and still have it match.

Ultimately, if I had to guess. I think we will have to stick with the way it is.

Tmcg2:
Rebased onto mater, still can't get that one function to match quite right. Will look into that soonish, have other stuff on my plate for now

MegaMech:
I took a look at the non match. Still no luck but I'm gonna keep trying.

Tmcg2:
The frustrating part is that if you do something like `struct_D_802F1F80 *temp_s0 = &D_802F1F80[arg3][arg2][arg1<<1];`, its nearly an exact match. The only incorrect part is the amount `arg1` gets shifted (which makes sense in context).

The register allocation then is expecting arg1 to be modified in-place in some capacity, but its unclear to me how. If my understanding of the `struct_D_802F1F80` type is correct, we shouldn't need to shift `arg1` in anyway, so I don't know what the original code could even look like.

`arg1` for this function is always a number between 0-7. As long as `D_802F1F80` itself has 8 elements in its lowest dimension, there shouldn't be a need for modifying `arg1` at all.

MegaMech (approved):

MegaMech:
Note that kart_dma.c func likely needs declare updated from struct to whatever. However, good enough for now.

PR #194, opened 2022-05-10
Added more actor types and some defines

Tmcg2:
Also adjusted the generic Actor struct to have some of its members
 have names that are generally reasonable for all types of actors


include/actor_types.h
```diff
@@ -5,6 +5,120 @@
...
+#define ACTOR_MRW_SIGN            0x17 // MRW -> Mario Raceway
+#define ACTOR_WS_SIGN             0x23 // WS -> Wario Stadium
+#define ACTOR_PADDLE_WHEEL        0x26 // Just the paddle wheel, or does it mean the entire boat?
+#define ACTOR_RR_CROSSING         0x27 // RR -> Railroad
+#define ACTOR_BLUE_SHELL          0x2A
+#define ACTOR_HAB_ITEM_BOX        0x2B // HAB -> Hot Air Baloon
+#define ACTOR_KIWANO_FRUIT        0x2D
+
+/*
+Used by the engine, coal car, and passenger car
```
MegaMech, line 31: prefer train engine or train locomotive here.  ~~coal car~~ tender.

include/actor_types.h
```diff
@@ -5,6 +5,120 @@
...
+*/
+struct train_car {
```
MegaMech, line 33: Do we really need a new struct for each object if they appear to use the generic actor struct?

Perhaps for modding purposes this makes sense?

include/actor_types.h
```diff
@@ -5,6 +5,120 @@
...
+    /* 0x08 */ f32 unk_08;
+    /* 0x0C */ f32 boundingBoxSize;
+    /* 0x10 */ Vec3s rot;
+    /* 0x16 */ s16 unk_16;
+    /* 0x18 */ Vec3f pos;
+    /* 0x24 */ Vec3f velocity;
+    /* 0x30 */ UnkActorInner unk30;
+}; // size = 0x70
+
+struct yv_egg {
```
MegaMech, line 74: Can we please do long names for everything. Other readers will see "yv_egg" out of context and not understand it.

include/actor_types.h
```diff
@@ -5,6 +5,120 @@
...
+#define ACTOR_BANANA              0x06
+#define ACTOR_GREEN_SHELL         0x07
+#define ACTOR_RED_SHELL           0x08
+#define ACTOR_YV_EGG              0x09 // YV -> Yoshi Valley
+#define ACTOR_PIRNHA_PLANT        0x0A
+#define ACTOR_ITEM_BOX            0x0C
+#define ACTOR_FAKE_ITEM_BOX       0x0D
+#define ACTOR_BANANA_BUNCH        0x0E
+#define ACTOR_TRAIN_ENGINE        0x0F
+#define ACTOR_TRAIN_COAL_CAR      0x10
```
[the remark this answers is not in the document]
MegaMech: Also, these appear to just be the wheels.

MegaMech:
I believe struct names should be: StructName

include/actor_types.h
```diff
@@ -5,6 +5,120 @@
...
+#define ACTOR_TRAIN_PASSENGER_CAR 0x11
+#define ACTOR_TRIPLE_GREEN_SHELL  0x15
+#define ACTOR_TRIPLE_RED_SHELL    0x16
+#define ACTOR_MRW_SIGN            0x17 // MRW -> Mario Raceway
+#define ACTOR_WS_SIGN             0x23 // WS -> Wario Stadium
+#define ACTOR_PADDLE_WHEEL        0x26 // Just the paddle wheel, or does it mean the entire boat?
```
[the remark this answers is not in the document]
Tmcg2: The struct associated with this type is definitely for both the boat and the wheel itself. The `boatRot` member controls the rotation of the boat itself, while `wheelRot` is for the paddle wheel itself*.

*Presumabaly, whenever I try locking the value of `wheelRot` I get a crash. Not sure why but it makes it hard to confirm whether that is truly the rotation of the wheel.

include/actor_types.h
Tmcg2: Maybe not? Most of the objects I've added structs for appear to have enough differences that a new struct type is useful for human readability (in my opinion at leat). Except maybe the FallingRock struct since that is a pretty "vanilla" object type compared to say, piranha plants.

include/actor_types.h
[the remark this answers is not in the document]
Tmcg2: No, these are definitely for the entire car. If you lock `rot[1]` for a train car in the desert map and then watch it go around the track, the car itself stops rotating to match the direction of movement (which can look pretty humorous). The `wheelRot` member by contrast is the rotation of the wheels on the car. Locking that value will make the wheels look static.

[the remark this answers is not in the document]
MegaMech: I guess my scientific method of testing is bunk. I think I know why I thought that. You're right.

include/actor_types.h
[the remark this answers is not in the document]
MegaMech: Okay, so I was only looking at the simple object func. And I guess the same ID gets used elsewhere as well. You're definitely right for both these scenarios.

include/actor_types.h
MegaMech: The piranha plants literally didn't match without its own struct.

Sounds reasonable to me, we'll go with your method.

include/actor_types.h
```diff
@@ -5,7 +5,121 @@
...
+#define ACTOR_TRAIN_TENDER        0x10
+#define ACTOR_TRAIN_PASSENGER_CAR 0x11
+#define ACTOR_TRIPLE_GREEN_SHELL  0x15
+#define ACTOR_TRIPLE_RED_SHELL    0x16
+#define ACTOR_MARIO_RACEWAY_SIGN  0x17
+#define ACTOR_WARIO_STADIUM_SIGN  0x23
+#define ACTOR_PADDLE_WHEEL_BOAT   0x26
+#define ACTOR_RAILROAD_CROSSING   0x27
+#define ACTOR_BLUE_SPINY_SHELL    0x2A
+#define ACTOR_HAB_ITEM_BOX        0x2B // HAB -> Hot Air Baloon
```
MegaMech, line 27: ACTOR_HOT_AIR_BALLOON

as well. Also, the comment is misspelled.

include/actor_types.h
```diff
@@ -5,7 +5,121 @@
...
+}; // size = 0x70
+
+struct PaddleWheelBoat {
+    /* 0x00 */ s16 type;
+    /* 0x02 */ s16 flags;
+    /* 0x04 */ s16 unk_04;
+    /* 0x06 */ s16 wheelRot;
+    /* 0x08 */ f32 unk_08;
+    /* 0x0C */ f32 unk_0C;
+    /* 0x10 */ Vec3s boatRot; // Weirdly, its boatRot[1] (Y rotation?) that changes the direction the boat faces
```
MegaMech, line 115: xyz. Isn't that pretty normal?

MegaMech (changes requested):
lgtm after my next comments is addressed.

include/actor_types.h
Tmcg2: I should have removed that comment. I think I know what's going on and its a matter of the rotation stuff being a little different than I expected. My presumption was that the `rot[Y]` value changing would mean the object rotates in the Y direction, but it appears that it actually means the that object rotates around the Y axis. So its entirely sensible I was just misunderstanding what the `rot` actually represented

MegaMech: Ah. rotAround? Might be q better term. rotTranslate?

MegaMech (changes requested):
Balloon has two Ls in it

Tmcg2:
Messed it up twice, third times the charm

MegaMech (approved):

PR #197, opened 2022-05-13
Give names to a couple vec3*_set functions

Tmcg2:


src/math_util.c
```diff
@@ -87,13 +87,17 @@ u32 func_802B5258(Vec3f arg0, Vec3s arg1) {
     return func_802B7830(temp_v1, temp_v2);
 }
 
-void load_giant_egg(Vec3f arg0, f32 arg1, f32 arg2, f32 arg3) {
+/*
+The difference between this and plain vec3f_copy is the return value.
+_alternate returns nothing while plain _copy returns a local variable address
+*/
+void vec3f_set_alternate(Vec3f arg0, f32 arg1, f32 arg2, f32 arg3) {
```
MegaMech, line 94: What about just `vec3f_set`? With no 'alternative'?

Tmcg2: There's a another function on line 131 (func_802B5320) that is identical to the `vec3f_set` function from the sm64  project. I opted to borrow the name from that project then. It is a bit of an odd choice since `vec3f_set` isn't actually used, instead `vec3f_set_alternate` is used.

I can swap their names, maybe give what is currently `vec3f_set` a new name like `vec3f_sm64`?

MegaMech: How does this sound:
`func_802B5320 -> UNUSED sm64_vec3f_set()`
`load_giant_egg` -> vec3f_set()
`func_802B52BC -> vec3s_set`

I think your explanation of the difference between the two would be unnecessary with it like this? Your choice whatever you think is most useful to the reader.

src/math_util.c
```diff
@@ -87,13 +87,17 @@ u32 func_802B5258(Vec3f arg0, Vec3s arg1) {
     return func_802B7830(temp_v1, temp_v2);
 }
 
-void load_giant_egg(Vec3f arg0, f32 arg1, f32 arg2, f32 arg3) {
+/*
+The difference between this and plain vec3f_copy is the return value.
+_alternate returns nothing while plain _copy returns a local variable address
```
MegaMech, line 92: This comment needs updating or removal

MegaMech (approved):

PR #202, opened 2022-05-16
Identify some Actor count variables

Tmcg2:
Also more accurately define the UknownRock struct. Some of the
 track specific data is spawning points for the "permanent"
 actors (stuff like falling rocks, shrubbery, item boxes, etc.)


MegaMech (approved):
I have my doubts regarding the for vs while loop. It also makes me wonder if the previous for loops where you have for (; ) // (blank ;) are actually whiles. But I dunno. Curious of other decompers have opinions

MegaMech:
Ya, can we keep it as a while? lmk if there's an issue with that. I'm guessing we have to keep that inner struct or that union

MegaMech (approved):

PR #206, opened 2022-05-19
Update gitignore & match more actor funcs

MegaMech:


src/code_80296A50.c
```diff
@@ -3466,44 +3435,29 @@ void func_8029E158(void) {
...
-//generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-? func_8029E7DC(); // extern
-extern u16 gNumActors;
 
-void func_8029E854(void *arg0) {
-    func_8029E7DC();
-    arg0->unk2 = 0;
-    arg0->unk0 = 0;
-    gNumActors = gNumActors - 1;
+// Deletes an actor?
```
Tmcg2, line 3454: pretty much, yeah. If shell or banana despawns due to getting hit this is how that actor is removed from the actor list.

src/code_80296A50.c
```diff
@@ -3017,60 +3006,40 @@ void place_all_item_boxes(struct ActorSpawnData *spawnData) {
...
-        phi_s0 = temp_s0;
-    } while (temp_s0 != 4);
+    //phi_s0 = 0;
+    for (i = 0; i < 4; i++) {
+        phi_s1 = &gPlayers[i];
+        //temp_v0 = *phi_s1;
+        if ((phi_s1->unk_000 & 0x4000) == 0) { continue; }
+        if ((phi_s1->unk_000 & 0x100) != 0) { continue; }
+
+        phi_s0 = func_8029EC88(sp64, sp50, sp58, (s16)45);
```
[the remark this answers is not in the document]
MegaMech: > arbitrary values from the stack
They do this a lot right?

Yea, order matters for the sp's. iirc m2c didn't place sp50 in the right spot. There was definitely some messing around plus I believe the continue statements were required here rather than `if () {// do stuff}`. Tricky stuff for sure.

src/code_80296A50.c
MegaMech: I think it's a multi step process for actor removal. This doesn't seem like enough code.

PR #211, opened 2022-05-25
Added some code I decompiled

RunTheCoins:
I wanted to contribute some code to this so yeah I attempted to do something


MegaMech:
I didn't have a thorough look yet. However, it appears you have decomped hand written asm files. Ido is a very... Primitive and specific compiler. There's a few commands it never uses that show up in hand written code. Therefore it will never match.

If this is the case, your PR may still be useful. Feel free to ask questions in Discord. Hand written asm is located in asm/ except for those bss files. The rest of the asm is in asm/non_matching/. Not that you can open any src/ file and remove the ifdefs and decomp that.

MegaMech:
thread9_crash_screen in crash_screen.c hasn't been decomped yet. So your decomp of that is usable.

func names are in this style: func_name_underline_between_words();
global symbols: gSymbolName
file only or local symbols: sSymbolName
(also known as static)

PR #213, opened 2022-05-25
Match some audio functions

Tmcg2:
Thank you Mech for figuring out what the stack management stuff was
 about.

And yes, apparently the best way to match those functions is to have
 an `if` statement that uses an unset local variable as its condition


src/audio/external.c
```diff
@@ -5526,38 +5527,23 @@ void func_800CA008(u8 arg0, u8 arg1) {
 GLOBAL_ASM("asm/non_matchings/audio/external/func_800CA008.s")
 #endif
 
-#ifdef MIPS_TO_C
-//generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-extern s8 D_800EA108;
-
 void func_800CA0A0(void) {
+    s32 pad;
+    if (pad){};
```
MegaMech, line 5532: Can we just do:
`if (pad);` with no `{}`?

MegaMech (changes requested):

src/audio/external.c
Tmcg2: Doesn't appear to match if  you do that. Without the `{};` the `if` statement will implicitly wrap around the actual meat of the function (since its a single line).

So we have to have the `{};` to make the `if` statement truly empty. I don't really understand why the compiler bothers to create stack space if its not going to bother including an actual break in the final asm, but it does so we have to have the `if` statement.

MegaMech (approved):

PR #217, opened 2022-06-11
Match code_802B0210.c

Tmcg2:
Required getting the .rodata for that file correct too.

Matched a couple other funtions here and there as well.


data/data_121DA0_2.s
```diff
@@ -5,42 +5,8 @@
...
-.word 0x00000000, 0x00000000
-
-glabel D_802B91A8
-.word 0x00000000, 0x00000000, 0x00000000
-
-glabel D_802B91B4
-.word 0x00000000, 0x00000000, 0x00000000, 0x0000000d
-.word 0x0000000d
+# I have no idea where this data belongs. It might be some initalized but unused variable?
+.word 0x0000000d, 0x0000000d
```
MegaMech, line 9: If the previous labels are initialized in code now, did you add this unused variable if that seems to work for now?

MegaMech (commented):
Other than my question and seeing if we can match this last func. ltgm

data/data_121DA0_2.s
Tmcg2: I don't think there's any direct reference to these values in the code.
D_802B91B4 only takes up 2 words worth of space (its a Vec3s, which due to compiler padding decisions takes up 8 bytes rather the minimum of 6)
Then there's 1 word of 0's that is really the result of file alignment from the linker.
And then there's the mystery double 0xD's.

MegaMech: Ah, yea if that's file pad then it would go in the next file. Looks good then!

MegaMech (approved):

PR #224, opened 2022-06-15
Add building on macOS

tehzz:
This PR adds in building for ARM macOS. It's based on my PR from the static-decomp repos (https://github.com/decompals/ido-static-recomp/pull/5). Allegedly the update does not break anything for x86 linux or mac, but it should be doubled checked.

I also ran into some issues with the line ending adjustments. It was removing CRLF bytes from the IDO binaries. So, I threw in a .gitattributes file that I had used in other projects (it's based on https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings).


MegaMech (approved):

MegaMech:
Thanks for the PR!

PR #246, opened 2022-07-18
More miscellaneous matches

Tmcg2:
And more mips_to_c refreshes


MegaMech:
Does the refreshing ignore `#define NON_MATCHING` for funcs that are already close to matching?

MegaMech (dismissed):

Tmcg2:
I do the refreshing by hand and in general I try to leave NON_MATCHING functions alone. No need to get rid of an existing attempt if I can't come up with something better (or, at least, that I think is better)

src/crash_screen.c
```diff
@@ -83,120 +83,38 @@ void crash_screen_draw_glyph(u16 *framebuffer, s32 x, s32 y, s32 glyph) {
...
+    s32 row;
+    s32 column;
+
+    for (var_s0 = 0x2E, var_s2 = 0x28; var_s0 > 0x2C; var_s0--, var_s2++) {
+        for (row = var_s2; row < var_s0; row++) {
+            for (column = var_s2; column < var_s0; column++) {
+                if ((var_s0 - var_s2) == 6) {
+                    framebuffer[(row * 320) + column] = RED_COLOUR;
+                } else {
+                    framebuffer[(row * 320) + column] = WHITE_COLOUR;
```
MegaMech, line 113: Can you update the func to align with our latest diff:
(change func prototype and `framebuffer[row * 320 + column] =`
https://decomp.me/scratch/pUilB

and then I'll merge.

MegaMech (changes requested):

src/crash_screen.c
Tmcg2: I think its currently up-to-date with that? Prototype seems right and the array access is as you commented.

MegaMech: Oh, my bad! I didn't look that closely apparently.

PR #251, opened 2022-08-14
Match a lot of functions, add header for objects

Tmcg2:
Also refreshed a TON of functions and created/added to several
 header files


src/code_80005FD0.h
```diff
@@ -4,7 +4,10 @@
 #include "common_structs.h"
 #include "vehicles.h"
 
+// FIXME: This header is missing some funciton prototypes
```
MegaMech, line 7: spelling; `function`

src/code_80057C60.c
```diff
@@ -5,9 +5,15 @@
 #include <variables.h>
 #include <config.h>
 #include <defines.h>
-#include "code_80057C60.h"
-#include "math_util_2.h"
 #include "math_util.h"
+#include "math_util_2.h"
+//#include "code_800431B0.h" <- FIXME: create this header file
```
MegaMech, line 10: maybe use todo as a keyword instead as I think people would tend to search that moreso than 'fixme'

MegaMech (changes requested):

MegaMech:
Other than those two things, lgtm.

MegaMech (approved):

PR #257, opened 2022-09-12
Change a variety of constants to defines

coco875:


src/memory.c
```diff
@@ -1437,7 +1437,7 @@ s32 load_course(s32 arg0) {
     } else {
         gHeapEndPtr = 0x8028DF00;
     }
-    set_segment_base_addr(9, func_802A7D70((s32) temp_v0_2->offsetRomStart, (s32) temp_v0_2->offsetRomEnd));
+    set_segment_base_addr(9, func_802A7D70(temp_a2, temp_v0->offsetRomEnd));
```
MegaMech, line 1440: Could you please undo this change?

src/menus.c
```diff
@@ -95,21 +95,24 @@ const s8 D_800F2B50[] = {0, 1, 2, 3, 3, 0, 0, 0};
 // set to D_8018EDF3, then that sets gPlayerCountSelection1
 const s8 D_800F2B58[] = {1, 2, 2, 3, 4, 0, 0, 0};
 
-const s8 D_800F2B60[5][3] = {
-    {1, 2, 1},
-    {1, 2, 1},
-    {0, 2, 2},
-    {0, 2, 0}, 
-    {0, 2, 0}
+// limit for each colone in one-two-three-four players selection
```
MegaMech, line 98: Do you mean column? And could you please capitalize the first letter of the sentence:
`Limit for each column...`

src/code_80027D00.c
```diff
@@ -6597,66 +6594,42 @@ void func_800381AC(Player *player, struct Controller *controller, u8 arg2) {
 GLOBAL_ASM("asm/non_matchings/code_80027D00/func_800381AC.s")
 #endif
 
-#ifdef MIPS_TO_C
```
MegaMech, line 6600: Removing this ifdef might be failing the build. (The output rom doesn't match the baserom)

Try running `./diff func_800382DC` (add -mw to auto run when you make changes).
This will show you the difference in assembly between the original game and the decompiled. To get OK the assembly and registers need to match 1:1.

If you don't have `diff` you'll need to run `git submodule update --init --recursive` (as shown in the readme).


You can also run `python3 first-diff.py` to see what the first difference in rom is.

MegaMech (changes requested):

MegaMech (changes requested):

MegaMech (approved):

PR #265, opened 2022-10-10
EU 1.0 OK

MegaMech:
Edit: Converted assets.json to yaml. Should be possible to add splat support with a bit of effort.

EU is not currently buildable. However, it appears quite similar to US.

Matched EU's main.c and added a splat yaml for EU.

*FIRST BUILD US*

Then build EU and work with it using the following commands.
`make -j VERSION=eu`
`python3 first-diff.py --eu`
`./diff <function> -eu`


MegaMech:
Not sure why we want this open other than easy access.

I feel like EU OK should just be new PR.

CoderStig (approved):

PR #268, opened 2022-11-01
Some matches for audio/synthesis

Tmcg2:
This required updating the Makefile to make the VERSION_US file
 actually gets used when compiling stuff.

Updated the abi.h file to match SM64, with some modifications to
 support the currently targeted version (US).


MegaMech (approved):

MegaMech:
Were you referencing SM64 EU? Just a heads up that SM64 US does not match mk64 but EU usually does.

Tmcg2:
Yes I meant SM64 EU. Good to know that EU generally matches the MK64 stuff. I tried matching some other stuff using the SH version but was having issues. I'll revisit it with EU instead

PR #273, opened 2022-11-16
Matches for audio/external

Tmcg2:
A lot of these matches come from observing the external audio
 stuff in SM64 and realizing that a couple of the structs
 in MK64 are just expanded versions of their SM64 counterparts.

Presumably all the expansions are for handling the doppler effect,
 although I can't say that with certainty


src/audio/external.c
```diff
@@ -1,9 +1,16 @@
...
 #include <defines.h>
+#include "camera.h"
+#include "math_util_2.h"
+#include "variables.h"
 #include "audio/external.h"
+#include "audio/load.h"
+#include "audio/data.h"
 
-GLOBAL_ASM("asm/non_matchings/audio/external/func_800C13F0.s")
+// REQUIRES void in the argument list to match properly
```
MegaMech, line 11: func(void) is normal for C.

src/audio/external.c
```diff
@@ -5568,21 +4891,17 @@ void func_800CA008(u8 arg0, u8 arg1) {
 GLOBAL_ASM("asm/non_matchings/audio/external/func_800CA008.s")
 #endif
 
-void func_800CA0A0(void) {
-    s32 pad;
-    if (pad){};
+// With -framepointer active, you CANNOT put void
+// in the argument list, causes a minor stack difference
+void func_800CA0A0() {
```
MegaMech, line 4896: Now that's weird...

MegaMech (approved):

src/audio/external.c
Tmcg2: Yeah, that's why I commented on the `func(void)` thing. Apparently there is a miniscule difference between `func(void)` and `func()`.
The first indicates that you cannot pass ANY arguments when invoking the function. `func(arg1, arg2, ...)` will cause a compiler error.
The second indicates that you can pass as many arguments as you want, its just that none of them are actually usable/reachable from within the function. So `func(arg1, arg2, ...)` will compile just fine.
IDO appears to make a further distinction between the two where `func(void)` leads to a slightly larger stack frame (at least when `-framepointer` is used. For some reason. Doesn't really make sense to me but that's what it looks like

PR #275, opened 2022-11-20
 Matches for audio/port_eu,external and fix lang problem

coco875:


src/audio/heap.c
```diff
@@ -604,77 +584,40 @@ void *get_bank_or_seq_inner(s32 poolIdx, s32 arg1, s32 bankId) {
 }
 
 #ifdef MIPS_TO_C
-//generated by m2c commit 0927f17aac197848d4ebdf0c6bbad74b01f0851c
+// error : `.rodata' referenced in section `.text' of build/us/src/audio/heap.o: defined in discarded section `.rodata' of build/us/src/audio/heap.o
+// inspired by func_eu_802e27e4_unused from SM64
```
MegaMech, line 588: Remove this comment

src/audio/heap.c
```diff
@@ -604,77 +584,40 @@ void *get_bank_or_seq_inner(s32 poolIdx, s32 arg1, s32 bankId) {
 }
 
 #ifdef MIPS_TO_C
```
MegaMech, line 586:
```suggestion
#ifdef NEEDS_RODATA
```

Makefile
```diff
@@ -13,6 +13,7 @@ default: all
 VERSION ?= us
 # If COMPARE is 1, check the output sha1sum when building 'all'
 COMPARE ?= 1
+export LANG := C
```
MegaMech, line 16: does the decomp.me decompile button still work after adding this line?
`./import src asm --decompme`

MegaMech (changes requested):

Makefile
coco875: I try and yes it still work

src/audio/heap.c
coco875: done

MegaMech (approved):

src/audio/port_eu.c
```diff
@@ -304,88 +300,54 @@ extern void *D_8019403C;
...
-#endif
 
-#ifdef MIPS_TO_C
-//generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-void func_800CBB48(? *); // extern
-
-void func_800CBB88(? arg1) {
-    func_800CBB48(&arg1);
+void func_800CBB88(u32 arg0, f32 arg1) {
+    func_800CBB48(arg0, (s32*) &arg1);
```
Tmcg2, line 319: If it works it works, but it seems really strange to typecast a float pointer to an int pointer

MegaMech: Yes please do `func_800CBB48(arg0, &arg1);` as that matches without the cast to s32

MegaMech: Actually, this is from sm64. It's fine as is.

Tmcg2: Interesting

MegaMech: I think the reasoning for this cast is for the programmer. So when you read it, it's like "Hey, this f32 gets cast to s32" rather than it being there for programmatic purposes.

PR #276, opened 2022-11-20
Matches in 80281fa0, render_courses, and create defines for sound cues.

MegaMech:
Added sounds.h which contains most sound cues. It's missing the tyre screech sounds.

Note that music there isn't really any music other than a few short sound cues.

Comments on naming appreciated.


data/data_121DA0.s
```diff
@@ -5,40 +5,40 @@
 
 .section .data
 
-glabel gFloatArray802B8790
```
Tmcg2, line 8: Any reason all this stuff is commented out instead of just deleting the file?

src/code_80281FA0.h
```diff
@@ -7,19 +7,19 @@
```
Tmcg2, line 7: Couldn't hurt to rename this to ceremony_and_credits.h

Also, is there a need to comment out a bunch of the forward declarations?l

MegaMech: I commented them out because I was changing function types and I would have to change it in .c and .h.

Will fix

Thanks for taking a look at the PR.

PR #279, opened 2022-11-26
Add a utility that aids in generating m2c output

Tmcg2:
Don't know if this would be useful to anybody else or not, but I figured it couldn't hurt to put it up for consideration

m2c_helper is a script I've been using to regenerate m2c output of functions as more and more stuff gets matched and/or added to header files.

```bash
m2c_helper.sh directory_name function_name
```
`directory_name` is the name of a directory under `asm/non_matchings`. So stuff like `code_80091750` or `audio/seqplayer`.
`function_name` is the name of the function you want to target. So stuff like `func_800721C0` or `find_unused_obj_index`

In practice you're meant to run the utility twice for the same function. The first time will create a file under `contexts/<directory_name>`. In its initial state it will look something like this:
```cpp
struct _m2c_stack_func_8009CBE4 {
    /* 0x00 */ char pad0[0x3C];
    /* 0x3C */ u32 *sp3C;                           /* inferred */
    /* 0x40 */ u32 *sp40;                           /* inferred */
    /* 0x44 */ char pad44[0x24];
};                                                  /* size = 0x68 */

extern ? D_8018E7AC;
extern ? D_8018E7B8;
extern ? D_8018E7D0;
extern ? D_8018E7E8;
extern ? D_8018E810;

void func_8009CBE4(s32 arg0, s32 arg1, s32 arg2) {
... rest of the function here ...
}
```
In practice I usually end up stripping this file down to just:
```cpp
void func_8009CBE4(newType *newarg0, s32 newarg1, otherType arg2);
```
You can opt the specify things in the m2c stack struct if you want, although I find that to be low value. But there's no reason not to have the option available. You can also add stuff to this file if you want. For example when working on functions in `code_80091750` its occasionally useful to add a line for `segmented_to_virtual_dupe` that returns a more specific type than `void*`.

After modifying the new context file file you re-run utility which will spit out the new disassembly to file specified by `DISSASEMBLY_FILE`.

The importing of data files only works if you apply the `m2c.patch` to m2c. There's some oddities with how stuff is included in some of our data files which m2c isn't expecting. The last hunk of the patch in particular is super specific to MK64.

The rest of the changes are related to making `m2ctx` usable. In the past I've been hand curating a `common.c` context file that had all the contents of the header files in the project. In the past I was unable to get anything useful out of `m2ctx` but in the past couple months someone (tehzz?) updated it and now it works fine. Only changes needed was adding newer headers to `common_includes.c` and making a tiny change to `bstring.h` to remove the `#ident` line. There's nothing wrong with that line but m2c won't handle it correctly, and I didn't want to have keep removing it from the resulting `ctx.c` file.


MegaMech:
I'm fine with adding this to the repo.

1) Can we have a more descriptive name of "m2c_helper"? (it's fine as is if not.
2) Can we a have a few more comments in the script files? Just general explanation of what it does is probably good enough.

What do you mean by?
> There's some oddities with how stuff is included in some of our data files which m2c isn't expecting

Perhaps it's easier to update the repo to be more in-line with what m2c expects?

Tmcg2:
The "oddities" in question have to do with `.incbin` lines. Using `kart_textures.s` as an example, all the `.incbin` lines only have the label name and the file to include. What they do not have is an offset and size for that include. I don't actually know much about how this works in practice, but my assumption is that you can have a really really big binary file and `.incbin` some `size` bytes starting at `offset` from the start, instead of needing to break the large bin file into smaller parts.

m2c assumes that those `offset` and `size` values are present on `.incbin` lines. So unfortunately we're kind of stuck having to patch m2c as I don't think going through and trying to add an `offset` and `size` for each and every `.incbin` line is really do-able.

I could push up the relevant parts of the patch as a PR to m2c itself, couldn't hurt to try. Still have to have the part where we explicitly skip `data/rsp.s`. I don't remember what's wrong with it, I just know that there's something about it that m2c doesn't like.

MegaMech:
There's only like 12k lines that would have to be upd... Yee.

Tmcg2:
I'm not great at naming things, so while I cannot come up with a better name than `m2c_helper` I'm sure some one else can. I am open to suggestions. Prior to this PR the script was actually 2 different scripts named `generate_context` and `generate_disassembly`. Naming things is not my forte.

MegaMech:
PRing the patch to the main repo should be fine. However, can you find a way to work around the data/rsp.s bit?

Tmcg2:
Sure, I'll look into it

MegaMech:
Someone stole your PR! :[

No worries though, you were credited as the author.

What's the issue with rsp.s?

Tmcg2:
Wait, really? That's interesting.

Anyway, I don't remember what the issue with rsp.s is and I've been busy with other stuff. I haven't forgotten about this PR, just been busy.

Tmcg2:
Okay, finally managed to get the issue for `rsp.s` to show up again. The issue is that that assembly file has both a .text and .data section. Normally this would not matter at all, however the .text section is including raw byte data rather than having a bunch of assembly code. This causes m2c to get a bit confused since it doesn't know how to decompile functions-as-bytes, it needs them as MIPS assembly.

This creates a lot of harmless but noisy messages in the m2c output, which is why I was ignoring the file entirely in my patch

This is the section in question:
```
.section .text, "ax"

# 0x800D8F70 D9B70-D9C40
glabel rspbootTextStart
.incbin "bin/lib/PR/boot/F3D_boot.bin"
glabel rspbootTextEnd

# 0x800D9040 D9C40-DB020
glabel gspF3DEXTextStart
.incbin "bin/lib/PR/f3d/F3DEX.bin"
glabel gspF3DEXTextEnd

 # 0x800DA420 DB020-DC430
glabel gspF3DLXTextStart
.incbin "bin/lib/PR/f3d/F3DLX.bin"
glabel gspF3DLXTextEnd

# 0x800DB830 DC430-DD0A0
glabel aspMainTextStart
.incbin "bin/lib/PR/audio/aspMain.bin"
glabel aspMainTextEnd 
```

Tmcg2:
Alright, updated the helper script with helpful comments and explanations, as requested.

Also found a way to exclude `rsp.s` that doesn't require patching m2c.

Going to work on putting up a PR for m2c itself to have the handling of MK64 style .incbin directives. We'll see what they say about it. If they accept it, great, no more need for a patch. If its declined then we're out of luck

Tmcg2:
Alright, after much churn I think this PR is ready to go
Note that m2c must be checked out to a commit at or after `ce5cdff12a6c07c112eda022a420ee4af721c5df` as that is the commit where 1 argument `.incbin` directives become supported.

MegaMech:
Yea, the m2c devs are pretty awesome. They certainly follow other decomps

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+
+#
+# m2c_helper.sh
+# Utility to aid in the generation of m2c output for specific functions
+#
+# example usages:
+#  m2c_helper.sh code_80091750 func_800AAC18
+#  m2c_helper.sh audio/load audio_init
+#
+# Be sure to set DISASEMBLY_FILE to a filename you are comfortable with.
```
MegaMech, line 11: DISASSEMBLY_FILE

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+#   Defaulted to function_disassembly.c
+#
+# Its expected that the utiltity will be ran twice for any given function.
```
MegaMech, line 14: `It's`, `utility`

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+# The first run will create a context file specific to the target function
+# The second run will create the full disassembly input,
+#   assuming the appropriate funciton specific context file found.
```
MegaMech, line 17: function

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+# Context for all the non-function-specific stuff (type definitions, function prototypes, etc.)
+COMMON_CONTEXT="ctx.c"
+# Extra options when generating the function specific context
+CONTEXT_M2C_OPTS="--stack-structs --zfill-constants"
+# Extra options when generating the function disassembly
+DISASEMBLY_M2C_OPTS="--zfill-constants"
+# Directories to search when a .incbin line is encountered in a data file
+INCBIN_OPTS="--incbin-dir ./expected/build/us --incbin-dir ./"
+# File that the final disassebmly output will be placed. Can be named whatever you want
+DISASEMBLY_FILE="function_disassembly.c"
```
MegaMech, line 52: DISASSEMBLY_FILE

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+for file in $DATA_FILES_EXCLUDE; do
+	DATA_FILES="${DATA_FILES//$file/}"
+done
+
+# Context for all the non-function-specific stuff (type definitions, function prototypes, etc.)
+COMMON_CONTEXT="ctx.c"
+# Extra options when generating the function specific context
+CONTEXT_M2C_OPTS="--stack-structs --zfill-constants"
+# Extra options when generating the function disassembly
+DISASEMBLY_M2C_OPTS="--zfill-constants"
```
MegaMech, line 48: DISASSEMBLY_M2C_OPTS

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+make_function_context () {
+	echo "$(${M2C_DIR}/m2c.py \
+		${ASM_DIR}/${directory}/${function}.s \
+		${DATA_FILES} \
+		${INCBIN_OPTS} \
+		${CONTEXT_M2C_OPTS} \
+		--context ${COMMON_CONTEXT})" > "${CONTEXT_DIR}/${directory}/${function}.c" 2>&1
+}
+
+make_function_disasembly () {
```
MegaMech, line 73: make_function_disassembly

m2c_helper.sh
```diff
@@ -0,0 +1,98 @@
...
+# example usages:
+#  m2c_helper.sh code_80091750 func_800AAC18
+#  m2c_helper.sh audio/load audio_init
+#
+# Be sure to set DISASEMBLY_FILE to a filename you are comfortable with.
+#   Defaulted to function_disassembly.c
+#
+# Its expected that the utiltity will be ran twice for any given function.
+# The first run will create a context file specific to the target function
+# The second run will create the full disassembly input,
```
MegaMech, line 16: I think this isn't worded the best as I find it difficult to understand.

What about?
`The second run outputs the full m2c`

MegaMech (changes requested):

m2c_helper.sh
Tmcg2: Oh, yeah, I messed that sentence up badly

m2c_helper.sh
```diff
@@ -45,11 +45,11 @@ COMMON_CONTEXT="ctx.c"
 # Extra options when generating the function specific context
 CONTEXT_M2C_OPTS="--stack-structs --zfill-constants"
 # Extra options when generating the function disassembly
-DISASEMBLY_M2C_OPTS="--zfill-constants"
+DISASSEMBLY_M2C_OPTS="--zfill-constants"
 # Directories to search when a .incbin line is encountered in a data file
 INCBIN_OPTS="--incbin-dir ./expected/build/us --incbin-dir ./"
 # File that the final disassebmly output will be placed. Can be named whatever you want
```
MegaMech, line 51: One more disassembly misspelling. Last one, I promise.

m2c_helper.sh
```diff
@@ -45,11 +45,11 @@ COMMON_CONTEXT="ctx.c"
...
-DISASEMBLY_FILE="function_disassembly.c"
+DISASSEMBLY_FILE="tyler.c"
```
MegaMech, line 52: Is tyler.c on purpose? (if it is it's fine)

MegaMech (changes requested):

m2c_helper.sh
Tmcg2: Nope, that was a mistake on my part. I name it after myself for my own purposes but for general use I want it named something else. There's no reason to use my name as the default.

PR #280, opened 2022-12-03
Complete matches in ceremony_and_credits

Tmcg2:
I think struct_D_802876D0 might be the "real" CinematicCamera
 struct, but I don't know for sure

I also replaced a bunch of Camera pointer args with
 struct_D_802876D0 pointers instead. Best that I can tell there
 are only a couple functions that use plain Camera pointers

Moved some type definitions into the header file


src/ceremony_and_credits.c
```diff
@@ -1752,17 +1560,17 @@ struct struct_80284AE8 {
...
-    case id:                                                                                              \
-        cutsceneDuration = cutscene[sCutsceneShot].duration;                                                  \
-        cutscene[sCutsceneShot].shot(camera);                                                                  \
+#define CUTSCENE(id, cutscene) \
+    case id: \
+        cutsceneDuration = cutscene[sCutsceneShot].duration; \
+        cutscene[sCutsceneShot].shot(camera); \
 
-    if (!camera->cutscene) { return; }
-    switch (camera->cutscene) {
```
MegaMech, line 1765: This naming is from sm64 and is likely accurate as the code is pretty similar.

However, this is probably something we can come back to later. Probably fine for now.

MegaMech (dismissed):

MegaMech:
Nice!

src/ceremony_and_credits.c
Tmcg2: If you want, I could rename `struct_D_802876D0 ` to `CinematicCamera` and carry the relevant member names too. Stuff like `pos`, `lookat`, and `cutscene`

MegaMech: Yea, if stuff seems relevant for sure. (Don't have to label everything, just whatever few makes sense).

I'll probably go back through again and see if I can clean it up a bit more (if there's any remaining left over unused structs).

CinematicCamera is probably not the best name. I think it's suitable for now unless you can think of something better.

Tmcg2:
Updated the struct name and some member names

MegaMech (approved):

PR #285, opened 2023-01-01
Couple matches and giving names to some textures

Tmcg2:
I also found a way to remove a bin2c invocation from the Makefile, although it will make referencing a couple of pictures a little more awkward. I'm willing to undo that change if requested


assets.json
```diff
@@ -12189,61 +12189,59 @@
...
-"textures/132B50_1ECD8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1ECD8"]}},
-"textures/132B50_1EED8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1EED8"]}},
-"textures/132B50_1F0D8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1F0D8"]}},
-"textures/132B50_1F2D8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1F2D8"]}},
-"textures/132B50_1F4D8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1F4D8"]}},
-"textures/132B50_1F6D8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1F6D8"]}},
-"textures/132B50_1F8D8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1F8D8"]}},
-"textures/132B50_1FAD8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1FAD8"]}},
-"textures/132B50_1FCD8_tlut.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1FCD8"]}},
+"textures/gTLUTItemWindowNone.rgba16.png": {"meta":{"dims":[16,16]}, "offsets": {"us":["0x132B50", "0x1DED8"]}},
```
MegaMech, line 12212: I'm wondering if gItemFrame would be a more accurate name than window.

MegaMech:
The reason for bin2c was that certain tluts were not producing matching output.

If you found a way to make it work that's great. Any idea what was going on there? iirc, multiple images referenced the same tlut and the exporter just wasn't calculating the output correctly.

assets.json
Tmcg2: Maybe? I don't know what to call it really and decided to just copy what the user manual says.

Tmcg2:
```
$(N64GRAPHICS) -i $(BUILD_DIR)/textures/132B50_1D6D8.rgba16.ci8.inc.c -g textures/132B50_1D6D8.rgba16.ci8.png -s u8 -w 32 -h 32 -f ci8 -c rgba16 -p textures/132B50_1B4D8_tlut_gen.rgba16.bin
$(N64GRAPHICS) -i $(BUILD_DIR)/textures/132B50_1DAD8.rgba16.ci8.inc.c -g textures/132B50_1DAD8.rgba16.ci8.png -s u8 -w 32 -h 32 -f ci8 -c rgba16 -p textures/132B50_1B4D8_tlut_gen.rgba16.bin
$(BIN2C) textures/132B50_1DAD8.rgba16.ci8.bin $(BUILD_DIR)/textures/132B50_1DAD8.rgba16.ci8
```
Causes `textures/132B50_1B4D8_tlut_gen.rgba16.bin` to be generated twice. Once based on `132B50_1D6D8.rgba16.ci8.png` (mini kart bomb) and once from `textures/132B50_1DAD8.rgba16.ci8.png` (question mark)

The n64graphics commands being used read like `Given a PNG, create a TLUT for the PNG and then convert the PNG into an equivalent .inc.c file based on the TLUT`, as opposed to `Given a PNG and a TLUT, convert the PNG into an equivalent .inc.c file`

As best as I was able to tell the problem was that the command for the question mark was working fine (the n64graphics utility isn't broken) but wasn't the correct command for what is needed.

There's a single TLUT for both the mini bomb kart and the question mark portrait and the only way to make n64graphics generate the correct `.inc.c` file for both images is to extract them as a single PNG and create a single combined `.inc.c`. If there were a way to tell n64 graphics "use this specific TLUT file" instead of "create a TLUT file", this probably wouldn't be an issue. But I know nothing about image to byte array conversions so that might be impossible. Someone with more knowledge on the subject would have to comment on that.

PR #287, opened 2023-01-17
Make the course models in a more Makefiley way

Tmcg2:
Required renaming one of the course directories to match the name
 used elsewhere in the project


Makefile
```diff
@@ -510,162 +505,14 @@ endif
...
-
-$(battle)/skyscraper/%.inc.mio0.o: courses/battle/skyscraper/%.inc.c
-	$(LD) -t -e 0 -Ttext=0F000000 -Map $(battle)/skyscraper/$*.elf.map -o $(battle)/skyscraper/$*.inc.elf $(battle)/skyscraper/$*.inc.o --no-check-sections
-	$(V)$(EXTRACT_DATA_FOR_MIO) $(battle)/skyscraper/$*.inc.elf $(battle)/skyscraper/$*.inc.bin
-	$(MIO0TOOL) -c $(battle)/skyscraper/$*.inc.bin $(battle)/skyscraper/$*.inc.mio0
-	printf ".include \"macros.inc\"\n\n.section .data\n\n.balign 4\n\n.incbin \"build/us/courses/battle/skyscraper/model.inc.mio0\"\n\n.balign 4\n\nglabel d_course_skyscraper_packed\n\n.incbin \"bin/course_skyscraper_packed.bin\"\n" > build/us/courses/battle/skyscraper/model.inc.mio0.s
-	$(AS) $(ASFLAGS) -o $(battle)/skyscraper/$*.inc.mio0.o $(battle)/skyscraper/$*.inc.mio0.s
+COURSE_MODEL_TARGETS := $(foreach dir,$(COURSE_DIRS),$(BUILD_DIR)/$(dir)/model.inc.mio0.o)
+
+$(COURSE_MODEL_TARGETS) : $(BUILD_DIR)/%/model.inc.mio0.o : %/model.inc.c
```
MegaMech, line 510: Given that we're making this bettery. I think it would be a good idea to add a comment that describes what goes on here. Perhaps:

```
# Elf the course data to include symbol addresses then convert course data to binary and compress to mio0. The mio0 file is converted to an object file so that the linker can link it.
```

MegaMech (changes requested):

Makefile
Tmcg2: Sure thing

MegaMech (approved):

PR #291, opened 2023-03-21
Move all ro/data for code_8001F980 out of data/*

Tmcg2:


MegaMech (approved):

MegaMech:
lmk when you are ready for merge.

Tmcg2:
This one should be good to go. I'll probably try to do something similar for some other files. Don't know how feasible it is to do this project wide but I bet there's at least one other file that this can be done to.

PR #292, opened 2023-03-22
Relocate rodata for some OS related function

Tmcg2:
Required moving a struct def to another file, and I renamed a
 struct to something nicer


MegaMech:
In oot __osHwIntTable has a length of ten.

MegaMech (approved):

PR #295, opened 2023-03-28
Massive data relocation

Tmcg2:
Mostly relocating rodata from data/* files to the appropriate
 asm/nonmatching/* files, although there is some plain data
 relocating to be found in here too.

Big change to note is the addition of the code_802AAA70.c file,
 split from memory.c. Based on some padding in the ro/data for
 memory it seems clear that a split is needed, although it is
 not clear that exact spot chosen for the split is correct.
 Still, it gets the file padding for the ro/data and text segments
 correct so it can't be too wrong.

Also included is an update to the asm_processor tool to support
 the .hword directive. I made a PR to the main repo of that
 project with an identical change and it was accepted.


src/skybox_and_splitscreen.c
```diff
@@ -584,22 +568,19 @@ void func_802A4D18(void) {
     }
     switch (gScreenModeSelection) {                 /* switch 1; irregular */
     case SCREEN_MODE_1P:                                         /* switch 1 */
-        D_80150148 = 1.333333f;
+        D_80150148 = 1.33333334f;
```
MegaMech, line 571: Is there a difference between `1.333333f` and `1.33333334f`?

MegaMech (approved):

MegaMech:
Also just a couple merge conflicts to resolve. lmk if you have any issues.

src/skybox_and_splitscreen.c
Tmcg2: IDO seems to think so. `1.3333f` is turned into a double that looks something like `3fccccc9` but in order to match it needs to be `3fccccca`.
I suspect that if I did something like `4.0f/3.0f` that it wouldn't look quite as strange (I think that matches, didn't test it though)

Tmcg2:
Rebased and fixed merge conflicts

MegaMech (approved):

PR #296, opened 2023-03-30
Update recomp.cpp

XDOneDude:
fixing for msys2 build


MegaMech:
```
#if defined(_WIN32) || defined(_WIN64)
#include <capstone/capstone.h>
#else
#include <capstone.h>
#endif
```

MegaMech:
Actually. Try this please

https://github.com/Emill/ido-static-recomp/pull/15

Absoluc4s:
An alternative I use for building this is WSL which I'm able to build this perfectly on.

MegaMech:
Abs, that's not relevant, this is about building on Windows. Linux already works

MegaMech:
Closing because the PR is stale. I think the real solution is to update to the latest version of recomp

I've opened #315 so the issue doesn't get forgotten.

PR #306, opened 2023-04-10
New asset extraction utility

Tmcg2:
Draft for a new extract.py utility. Motivated by the quirks of
 how kart animation frames are stored in the ROM.

Long story short: each frame has a palette that is actually
 composed of 2 other, incomplete palettes. As such, the existent
 extract.py isn't really able to handle them.


MegaMech:
This looks like a great idea! Let me know if you need any help.

palette_extract.py or similar would make good names.

Tmcg2:
As previously mentioned, the kart animation frames are ci8 images whose palettes are a combination of 2 other palettes. The top 3/4's of the palette is for the kart body, while the bottom 1/4 is for the wheels. This presents a pretty annoying issue when attempting to create the inc.c/bin file for the frame. If a color is found in both the kart body and wheel parts of the palette the wheel indexes will get screwed up.
Below are instructions on how to create the files necessary to see the issue
First apply this patch to the new extraction utility. It causes the combined palette to be saved
```diff
diff --git a/new_extract_temp_name.py b/new_extract_temp_name.py
index cc0b8c9..a316d32 100644
--- a/new_extract_temp_name.py
+++ b/new_extract_temp_name.py
@@ -146,7 +146,8 @@ def export_image(baserom, asset, asset_list):
     ]

     if asset["type"] in ("ci4", "ci8"):
-        palette_file = tempfile.NamedTemporaryFile(mode="w+b", prefix="palette_")
+        #palette_file = tempfile.NamedTemporaryFile(mode="w+b", prefix="palette_")
+        palette_file = open("./temp_assets/bowser/palettes/bowser_kart_frame0000_pal.bin", "w+b")
         # This is INCREDIBLY DUMB, but it ensures that the palettes are found
         # and stitched in the same order they're specified in the 'tluts' key
         for tlut_name in asset["tluts"]:
```
Then run the following commands
```console
./new_extract_temp_name.py baserom.us.z64 ./temp_assets/big_issue_example.json
./tools/n64graphics -e ./temp_assets/bowser/palettes/bowser_kart_frame0000_pal.bin -g ./temp_assets/bowser/palettes/bowser_kart_frame0000_pal.png -w 16 -h 16 -f rgba16
./tools/n64graphics -Z ./temp_assets/bowser/bowser_kart_frame0000.bin -g ./temp_assets/bowser/bowser_kart_frame0000.ci8.png -w 64 -h 64 -f ci8 -c rgba16 -p ./temp_assets/bowser/palettes/bowser_kart_frame0000_pal.png
./tools/mio0 -c ./temp_assets/bowser/bowser_kart_frame0000.bin ./temp_assets/bowser/bowser_kart_frame0000.mio0
```
`./temp_assets/bowser/bowser_kart_frame0000.bin` can be compared to `./bin/karts/7_bowser/kart_000_57DE70.bin`
`./temp_assets/bowser/bowser_kart_frame0000.mio0` can be compared to `./bin/karts/7_bowser/kart_000_57DE70.mio0`
The primary difference is most notable in the first comparison. There are several entries for `c2` in the desired bin file, but in the produced bin file those values are `74`. This is because the color at index `c2` in the palette is also found at index `74` and due to how `n64graphics` works `74` will always be chosen when producing the bin file.
I have no idea how to solve this problem right now.

Tmcg2:
I've though of 1 solution, although its really really bad.

When extracting a kart animation frame, we could make a companion file that indicates whether a given pixel belongs to the "kart" or the "wheels". When extracting the frame in its `ci8` form we could consider each index value and check: is it between `0x00` and `0xbf`? If so, put a `0` in the companion file (pixel is in the kart palette). Otherwise, put a `1` in the companion file (pixel is in the wheel palette).

Then, we add yet another mode the `n64graphics` that takes a image (PNG), a palette (PNG), and  a companion file as input. When searching for a given color from the image in the provided palette, it first consults the companion file. If the companion file has an entry of `1` for that pixel it starts the search in the palette as index `0xc0` instead of `0x00`.

That would work (I think), but its ugly. But I also haven't thought of anything better yet.

MegaMech:
I have an idea.

Since you're creating a json file which contains meta-data. We should have a property for "type": wheel | kart.

If it's a wheel it does a reverse lookup of the palette. Kart does a normal lookup.

ie. The reverse lookup starts at the end of the file and goes up.

MegaMech:
Regarding .m64, sm64 has some tools to for it already I think. That will be a big can of worms on its own.

Tmcg2:
Yeah I've been looking at the `seq64` repo. I most of the `.m64` files converted to midi files using that. `seq_00` doesn't convert since, I think, that's a bank of sound effects (some documentation I saw somewhere says sound effects are in the first bank in all(?) games)
seq_08, _09, and _10 aren't converting right though, not really sure what's up with that.

Tmcg2:
Implemented some changes to `n64graphics` that allows the `-Z` mode to use stitched palettes correctly. Long story short, you need to provide a "wheel mask" file (produced by the new extract utility) that indicates whether a given pixel in an image belongs to the "kart" or the "wheel"

Tmcg2:
All right, got this to a state I'm happy with it.

The new extraction tool is under the tools directory and is currently used to extract the kart textures. Unfortunately, I couldn't find a way to get the Makefile to play nice, so now the build process need a slight change. You'll have to run `make extract_assets` first. Once that's done you should be able to do plain `make` to do the build proper. It shouldn't be necessary to re-run `make extract_assets` unless you: modify the `kart.json` files or do `make distclean`.

I feel like you'd have to change how the Jenkins stuff works to get the build-on-push hook to work.

As a downside, the `extract_assets` step takes an annoyingly long time (upwards of 10 minutes). I don't really know what's going on with that, I assume that its combination of me using Python and the tool not being in an efficient manner. But, since it only needs to be ran once I don't think it'll be a huge problem.

MegaMech:
This looks good.

I've been pondering the ten minute build time. I believe the issue is because we're loading and saving thousands of files multiple times. Windows file system is very slow.
Just one file requires:
1) Load file and create a tmp copy
2) Launch n64graphics/mio0 programs
3) Convert file and save.

This is a lot of processing for thousands of files.

Whereas Python is pretty slow including sub_process calls, I do think that's not necessarily the bottleneck.

A more performant solution would consist of running the C program once, loading all the files into ram, converting them, and then saving them. Even faster would be if we saved them all as one big singlular file or multiple big files.

I'll ponder about how this might be done a some more.

Tmcg2:
Figured out how to get the build working without needing to do 2 make calls. Requires having some large hard coded lists of file names. But I don't anticipate those lists changing ever, so I don't think the hard codedness will be a future problem

MegaMech:
That's a great solution!

Tmcg2:
I didn't mean to include the changes to the packed_displaylist or vertices_count files. Don't know why they keep showing up in my work tree, don't know if that's a me problem or not.

MegaMech:
If definitions in those generated files get removed then we'll have build issues. The tool that generates those just barely works.

I'll try experimenting with this PR in the next couple days

MegaMech:
Did I skip a step?

make: *** No rule to make target 'assets/finish_line_banner.json', needed by 'assets/finish_line_banner/.export'.  Stop.

Tmcg2:
Probably not? I feel like I saw that error, it might be a fix I forgot to push up. I'll check

MegaMech:
If you comment out line 333 of makefile for now that should prevent the segment headers tool from running.
`$(DLSYMGEN)`

MegaMech:
The compile time is about one minute and twenty-two seconds. Most of that is recomp.

I don't believe this PR effects compile time all that much. However deletion time is ~2 minutes which I believe is a substantial increase.

There's a makefile bug requiring running make twice.

The directory structures and naming are really good.

assets/finish_line_banner.json
```diff
@@ -0,0 +1,11 @@
+{
+"gTLUTFinishLineBanner":     {"output_dir": "finish_line_banner", "rom_offset": "0x132B50", "block_offset": "0x0", "width": 16, "height": 16, "type": "rgba16"},
+"gTexutreFinishLineBanner1": {"output_dir": "finish_line_banner", "rom_offset": "0x696BAC", "width": 64, "height": 32, "type": "ci8", "tlut": "gTLUTFinishLineBanner"},
```
MegaMech, line 3: gTexture

include/packed_displaylist_symbols_gen.h
MegaMech, line 1: Discard this change if you can.

If need be I can save a copy locally and merge a new one after merging your PR.

Same goes for vertice_count_gen.h

tools/new_extract_assets.py
MegaMech, line 1: lgtm. Does this replace the old extract_assets? Or does it work in-tandem with it? If so, rename the old one to `unused_old_extract_assets.py` and we'll let it sit around for a bit in-case the code comes in-use as a reference in the future.

MegaMech:
Lastly the readme's Project Structure section needs updating.

I can do that if you don't want to.

tools/new_extract_assets.py
Tmcg2: For now both are needed. I intend to try and move all asset extraction to the new one eventually but for now the old one is still doing its job.

include/packed_displaylist_symbols_gen.h
Tmcg2: Can do

assets/finish_line_banner.json
Tmcg2: Good find

Tmcg2:
Man I somehow really messed up the fixing of the gTexutre thing. Should be good now

Tmcg2:
I'd also hold off on the Project Structure changes for now, I have some other changes for a different PR that moves a few things around. This is one of those things that the more I look at it the more I see room for improvements. Shouldn't be a huge change though.

Tmcg2:
The PR is good to go though if you're happy with it. I don't intend to touch it further, other changes I'm thinking are better off being a separate PR

tools/new_extract_assets.py
MegaMech: If both files do different things, it's okay to leave them separated too, if it makes sense to do that.

mk64.ld
```diff
@@ -454,7 +454,14 @@ SECTIONS
...
-      BUILD_DIR/data/kart_textures.o(.data);
+      BUILD_DIR/data/karts/luigi_kart.o(.data);
+      BUILD_DIR/data/karts/mario_kart.o(.data);
+      BUILD_DIR/data/karts/yoshi_kart.o(.data);
+      BUILD_DIR/data/karts/peach_kart.o(.data);
+      BUILD_DIR/data/karts/wario_kart.o(.data);
+      BUILD_DIR/data/karts/toad_kart.o(.data);
+      BUILD_DIR/data/karts/donkeykong_kart.o(.data);
+      BUILD_DIR/data/karts/bowser_kart.o(.data);
       BUILD_DIR/courses/staff_ghosts.inc.o(.data);
```
MegaMech, line 465: Do you think it might make sense to move staff_ghosts out of courses? (maybe not for this PR just wondering what you think)

Or does it still seem sensical for it to stay?

I have it this way cause the individual staff ghost file is in its relevant course folder.

Makefile
```diff
@@ -214,7 +220,7 @@ endif
 clean:
 	$(RM) -r $(BUILD_DIR)
 
-distclean:
+distclean: distclean_kart_assets
```
MegaMech, line 223: I'm going to test distclean before I merge.

MegaMech (dismissed):

mk64.ld
Tmcg2: I honestly don't know. I'd have to look at the file to see what that stuff is really doing. If its the code that handles "playing" the ghost in time trial mode it might make sense to move it into `src`. But I think you've touched that stuff the most so you're the one who'd know best.

Makefile
Tmcg2: Sure thing. All it SHOULD do is clean out the png/bin/mio0 file from the assets folder. Regular `clean` cleans out the build directory. I forget what all `distclean` does but it includes deleting the textures the original `extract_assets` exported. So `distclean_assets` is intended to be a super target that allows you to clean out the assets extracted by the new utiltity.

In the future I plan to add stuff like `distclean_lakitu`, `distclean_karts`, etc to the appropriate `.mk` file so you could target a specific group of assets. For now though the "burn it all" option is the only one available

mk64.ld
MegaMech: It's just the controller data for the staff ghost. Like A button, left, right, etc. The actual code is in src/staff_ghosts.c

Makefile
MegaMech: Should the distclean_kart_assets part be removed? Or did I do something wrong?
```
make distclean
make: *** No rule to make target 'distclean_kart_assets', needed by 'distclean'.  Stop.
```
Perhaps you meant to put `$(MAKE) distclean_assets`?

For distclean_assets. Can we add make clean?
```
distclean_assets:
	rm -rf $(ASSET_DIRECTORIES)
	$(MAKE) clean
```
This will help the user-experience. If you build after removing the assets it will error due to not realizing it needs to rebuild everything which could be confusing.

MegaMech:
This PR breaks parallelism for higher than 5 cores. However, 5 cores is plenty enough so I'm not worried. I can look into that a different time.

Makefile
MegaMech: Also,

Please comment out the call to `$(DLSYMGEN)` and `$(MODELSYMGEN)`

Tmcg2: No, I messed up here. It ought to be
```
distclean: distclean_assets
```
so that doing plain `distclean` also runs `distclean_assets`. `distclean_kart_assets` is a leftover from the PRs earlier life.

I don't want to put a subcall to `make clean` in `distclean_assets`, its intended to be its own thing, only cleaning out the assets. If someone wants to do a full clean they should run plain `distclean`, while `distclean_assets` is for cleaning out the asset stuff specifically.

Tmcg2: Why do DLSYMGEN and MODELSYMGEN need to be commented out?

MegaMech: Sounds good!

Because the DLSYMGEN and MODELSYMGEN scripts have bugs and we don't need them doing anything unless the game gets modified. Without them running we don't have to deal with broken header files.

PR #324, opened 2023-07-01
Give names to the item curves

Tmcg2:
Also use actual item names instead of raw numbers


src/code_80071F00.c
```diff
@@ -3788,10 +3788,10 @@ void func_8007AC9C(s32 playerId) {
...
 extern u8 gControllerRandom;
 extern s32 gRaceFrameCounter;
-extern s8 D_0D008B14[];
-extern s8 D_0D008470[];
-extern s8 D_0D008858[];
-extern s8 D_0D008984[];
+extern s8 gBattleItemCurve[];
+extern s8 gGPCpuItemCurve[];
+extern s8 gVS3pItemCurve[];
+extern s8 gVS4pItemCurve[];
```
MegaMech, line 3794: I'm not a fan of gGPCpu as it could be confused with GPC pu rather than GP CPU.

Do you think gGrandPrixCPUCurve would be okay? CPU is a very common abbreviation. So I think that should be fine.
Also gVersusThreePlayerItemCurve and gVersusFourPlayerItemCurve.

I know GP, VS, and TT, should be obvious enough (I think TT could get confusing). Readers not familiar with mk64 might struggle to read the code-base with these as abbreviations. I think there's a fair argument for 1p, 2p, 3p, 4p or p1, p2, p3, p4. But I think that could use a bit of discussion to come up with an approach that could be used consistently.

Ie. The code-base could end up with Player1 PlayerOne p1, etc. 1p and p1 could get confused. But if we're consistent with the usage it should be okay.

src/common_textures.inc.c
```diff
@@ -1748,246 +1749,339 @@ Gfx D_0D008138[] = {
...
+        ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,             ITEM_STAR,
+        ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,
+        ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,  ITEM_TRIPLE_MUSHROOM,
+        ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,   ITEM_SUPER_MUSHROOM,
+    },
 };
 
 // Versus mode four player
```
MegaMech, line 2019: Can you remove the comments now that we have suitable names?

MegaMech (changes requested):

src/code_80071F00.c
Tmcg2: Yeah, that sounds fine

MegaMech (approved):

PR #332, opened 2023-07-08
Migrate rodata for code_80005FD0 into asm files

Tmcg2:


src/code_80005FD0.c
```diff
@@ -17,6 +17,343 @@
 #include "hud_renderer.h"
 #include "code_80057C60.h"
 
+// Strings, presented by google translate!
+// Note that these are EUC-JP encoded, see:
```
MegaMech, line 21: Oh no, not more EUC-JP.

Every time I accidentally save over credits.c, I rage quit.

src/code_80005FD0.c
```diff
@@ -17,6 +17,343 @@
...
+// Direct BOM(%d) (%7.2f, %7.2f, %7.2f)
+char *D_800EBC24 = "ľ�ܻ����BOM(%d) (%7.2f, %7.2f, %7.2f) \n";
+char *D_800EBC50 = "BOM HIT CHECK\n";
+char *D_800EBC60 = "BOM HIT !!!!! (%d)\n";
+// BOM standby
+char *D_800EBC74 = "BOM�Ե�\n";
+char *D_800EBC80 = "RESULT BOM area(%d)\n";
+// BOM dropped.
+char *D_800EBC98 = "BOM �� ����ޤ�����\n";
+// Tortoise fire pillar SET failed (TABLE IS FULL)
```
MegaMech, line 78: Tortoise fire pillar? That sounds invigorating.

MegaMech (approved):

src/code_80005FD0.c
Tmcg2: Yep, its super annoying. And, unfortunately, this will not be the last file to have it. code_80091750 is going to need it too :(

MegaMech (approved):

MegaMech:
In the future converting the file to EU-JAP should be part of the build process. This way saving the file as US doesn't matter.

PR #338, opened 2023-07-09
Move some data into kart_dma.c

Tmcg2:


MegaMech (approved):
Is it worth considering putting these definitions in a separate file such as:
kart_textures.c
kart_dma.c

I think it could still match... Maybe?

PR #339, opened 2023-07-10
Possible Generator for course_data and new mio0 extractor

MegaMech:
This isn't necessarily going to make it into the repo. But if these changes seem good. They will likely require some massive changes to the build system. (well, not the course generator part) but the mio0 extractor would.

### New mio0 extractor

Edit: Removed json dependency. Uses tab delimited instead.

Every compressed asset requires running the mio0 program from Python. That's a lot of fopens and freads and fwrites.
Python is very expensive for shell calls too.

This new extractor links in with libmio0. It extracts all mio0 files into a buffer and writes that buffer as one big .bin file.

It also saves a ~~json~~ text delimitted file which links the real offsets to the new offsets. A future extract_assets program could process the extracted file without need to update assets.json.

This means a single fopen and fwrite. In the future, the baserom_extractor could pass the buffer straight to n64graphics or the asset extractor.

### Course Data Generator in Go 1.20.5

Course data layout defined using yaml.

Here's an example of what a generator in go would look like. There's still some cleaning up that could be done. But I think this is decent code and I really like how it uses mk64's Vtx struct and that the string concatenation looks pretty close to what you would actually see in the file.

Let me know if you have any thoughts.

Right now it only generates Vtx, but it's really easy to add more types.

There's two .go files.
main.go -> takes of care of cli, opening & writing files, etc.
generators.go -> processes the data and concatenates strings.

Compile using: `go build -o gen`
Run using: `./gen -yaml=courses.yaml -baserom=../../baserom.us.z64`

Example of what it outputs right now:
```
/* @file course generated by static extractor */

some gfx
some gfx
some gfx
some gfx

Vtx some_var[] = {
    {{ 0, 10277, 4096}, 22, { 364,4129}, { 0x00, 0xC3, 0x00, 0x19}},
    {{ 0, 10277, 4096}, 22, { 364,4129}, { 0x00, 0xC3, 0x00, 0x19}},
    {{ 25, -28744, 52}, 15385, { -32746,10041}, { 0x30, 0x50, 0x03, 0x19}},
    {{ 793, 4129, -31666}, 0, { 15361,16384}, { 0x44, 0x8E, 0x40, 0x00}},
    {{ 16384, 0, 20498}, 17546, { 12288,18048}, { 0x41, 0x20, 0x44, 0x81}},
    {{ 17537, 16384, 18048}, 13472, { 17922,-28029}, { 0x46, 0x0A, 0x24, 0x00}},
    {{ 9216, 17928, -32381}, 17920, { 13453,17413}, { 0x90, 0x00, 0x00, 0x00}},
    {{ 0, 0, 5}, 31744, { 15,11267}, { 0x00, 0xE0, 0x20, 0x25}},
    {{ 8229, -20574, 60}, 42917, { 148,3072}, { 0xDA, 0x03, 0xAF, 0xA8}},
    {{ -20568, 32, -28766}, 60, { -30811,148}, { 0x8F, 0xAC, 0x00, 0x28}},
    {{ 40, 9217, 1}, 42053, { 0,-29288}, { 0x00, 0x00, 0x8F, 0xA8}},
    {{ -28760, 32, 4865}, 13, { 0,0}, { 0x8F, 0xB9, 0x00, 0x2C}},
};


eof
```


tools/generator/generators.go
```diff
@@ -0,0 +1,76 @@
...
+        x := fmt.Sprint(vtx.Ob[0])
+        y := fmt.Sprint(vtx.Ob[1])
+        z := fmt.Sprint(vtx.Ob[2])
+        flag := fmt.Sprint(vtx.Flag)
+        tc1 := fmt.Sprint(vtx.Tc[0])
+        tc2 := fmt.Sprint(vtx.Tc[1])
+        cn1 := fmt.Sprintf("0x%02X", vtx.Cn[0])
+        cn2 := fmt.Sprintf("0x%02X", vtx.Cn[1])
+        cn3 := fmt.Sprintf("0x%02X", vtx.Cn[2])
+        cn4 := fmt.Sprintf("0x%02X", vtx.Cn[3])
```
MegaMech, line 45: Probably more performant to write all of the Vtx into a struct var and then print that.

PR #341, opened 2023-07-12
Throwing my hat in the ring concerning the course data extraction stuff

Tmcg2:
This isn't meant to be competition with the existing proposal, rather
 its intended to be a comparison to see how it could be done in
 Python if desired.

The mio0_decode stuff could be migrated into `new_extract_assets.py`
 to potentially speed it up a little bit too, since it would cut
 down on the number of files it creates and reads from.

### DIFF
--- a/tools/vtx_gfx_extractor.py
+++ b/tools/vtx_gfx_extractor.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+
+import sys
+import struct
+import tempfile
+import subprocess
+import ctypes
+
+# libmio0.so has to compiled be like so:
+# gcc -o libmio0.so libmio0.c -shared -fPIC
+_mio0_decode = ctypes.CDLL("tools/libmio0.so")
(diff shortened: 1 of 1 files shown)


Tmcg2:
In its current state the utility will read extract `d_course_mario_raceway_piranha_plant_model` to `d_course_mario_raceway_dl_tree` from the Mario Raceway course data.

I don't know how we'd extend what I've currently got to accommodate stuff we `#include` (so stuff like TLUTs and textures), but add extractions for waypoints, actor spawns and other non vtx/gfx data, but I wouldn't expect it to be anything too crazy

tools/vtx_gfx_extractor.py
```diff
@@ -0,0 +1,100 @@
...
+
+import sys
+import struct
+import tempfile
+import subprocess
+import ctypes
+
+# libmio0.so has to compiled be like so:
+# gcc -o libmio0.so libmio0.c -shared -fPIC
+_mio0_decode = ctypes.CDLL("tools/libmio0.so")
```
Tmcg2, line 11: I guess `_mio0_decode` would be better called `_libmio0`, but that's a small detail

PR #352, opened 2023-08-06
Some mathes and some non-matchings in code_80057C60

Tmcg2:


MegaMech (dismissed):

Tmcg2:
Migrated the plain data for code_80057C60 into the appropriate `asm` files. So now there's more matches and some further extra changes to support the new linking that's needed

MegaMech (approved):

Issue #345, opened 2023-07-15
EUC-JP encoding


MegaMech:
Apparently sm64 saves files as US. And the build system converts the necessary files to EUC-JP

Any reason that solution isn't relevant here?

progress.py and totalBadge will be inaccurate until this is solved.

Priority: 0

MegaMech:
Fixed in #354

PR #353, opened 2023-08-06
Fix progress.py after euc-jp fix

MegaMech:
Addresses #345 

textconv taken from sm64. Likely used instead of iconv for portability.


Tmcg2:
I was actually trying to accomplish the same thing recently, lol.
I've been trying to use `iconv` to do the conversion though

MegaMech:
> I was actually trying to accomplish the same thing recently, lol. I've been trying to use `iconv` to do the conversion though

I think the progress.py percentage is off. But whatever.

Tmcg2:
iconv is being difficult, I'll keep working on it if only to satisfy my own curiosity. If your stuff works I'd just go forward with it

MegaMech:
Closing because #354 is a working solution.

PR #354, opened 2023-08-06
Run iconv as part of build

Tmcg2:
That way files that need to be in EUC-JP encoding can still be
 worked on in UTF-8


Tmcg2:
Note that as of now this PR doesn't work, something about `code_80005FD0` is breaking after the conversion into EUC-JP, I get spammed with some type of syntax error. Not sure what the source is yes
```
cfe: Error: build/us/src/code_80005FD0.jp.c, line 510: Syntax Error
 GLOBAL_ASM("asm/non_matchings/code_80005FD0/func_800065D0.s")
```

Tmcg2:
Digging into it, it seems like  IDO cc may not play nice with EUC-JP encoded files. It's handling credits.c okay but not code_80005FD0.c

Tmcg2:
I believe the issue was based on a misunderstanding on my part. I've always interpreted `GLOBAL_ASM` as something like a macro, but really its more like a special marker in the file. That marker is searched for by the `asm_processor` tool, which then replaces the `GLOBAL_ASM` line with the contents of the indicated `.s` file.

Since its not a macro, IDO `cc` was unable to interpret it, which was the source of all the error messages I was seeing. The solution was to pass the generated `jp.c` files through `asm_processor`, just like like we do for the regular `.c` files.

Tmcg2:
I guess its worth pointing out that this change creates an implicit requirement of having `iconv` installed. I didn't have to explicitly install it on an Ubuntu WSL install, but I don't know if other *nix flavors with have it installed by default.

Makefile
```diff
@@ -218,14 +218,20 @@ include $(MAKEFILE_SPLIT)
 
 COURSE_ASM_FILES := $(wildcard courses/*/*/packed.s)
 
-C_FILES := $(foreach dir,$(SRC_DIRS),$(wildcard $(dir)/*.c))
+# These are files that need to be encoded into EUC-JP in order for the ROM to match
+# We filter them out from the regular C_FILES since we don't need nor want the
+# UTF-8 versions getting compiled
+EUC_JP_FILES := src/credits.c src/code_80005FD0.c
+C_FILES := $(filter-out $(EUC_JP_FILES),$(foreach dir,$(SRC_DIRS),$(wildcard $(dir)/*.c)))
 S_FILES := $(foreach dir,$(ASM_DIRS),$(wildcard $(dir)/*.s)) $(COURSE_ASM_FILES)
```
MegaMech, line 226: This seems unnecessarily complicated.

What's wrong with compiling the utf8 verisons? They aren't linked. I don't see any issue with compiling them if it keeps the makefile a tad cleaner.

Thoughts?

Please update the readme to include iconv as a dependency and perhaps note that it generally comes packaged with gnu/linux.

MegaMech (changes requested):

Makefile
Tmcg2: Nothing really wrong with it, just wasted work (not that they take much time to compile of course). I'll remove the filter out though, no big deal.

As for the readme, I've done some googling and it seems like `inconv` is just part of GNU C library. Presumably any system that has stuff like GCC installed is going to naturally have `iconv`.

MegaMech (approved):

MegaMech:
Addresses #345

PR #355, opened 2023-08-07
small changes. One matching func, Two nonmatching funcs

TommiRommi:


include/macros.h
```diff
@@ -64,4 +64,6 @@
 // aligns an address to the next 16 bytes
 #define ALIGN16(val) (((val) + 0xF) & ~0xF)
 
+#define SQUARE(x) ((x) * (x))
```
MegaMech, line 67: I'm not sure if this is the right header for this.

How do you feel about putting it in mk64.h instead?

MegaMech (changes requested):

MegaMech:
Notice how your PR contains all the commits of the previous PR? This is because you PR'd twice from the same branch.

After your PR gets merged you should try the following steps:
1) Sync/update your master branch on your fork
2) Create a new branch
3) Re-clone that
4) Start making changes
5) Push those changes and PR.

This should prevent that from happening again.

Alternatively, you can learn how to re-base. But the above is the easiest approach.

include/macros.h
TommiRommi: Ok

MegaMech (approved):

MegaMech (approved):

PR #372, opened 2023-08-15
Push more matching! Primarily audio

ItzSwirlz:


src/audio/port_eu.c
```diff
@@ -298,30 +298,24 @@ void func_800CBBE8(u32 arg0, s8 arg1) {
...
 }
 
-#ifdef MIPS_TO_C
-// link of progress : https://decomp.me/scratch/1N93O
-extern OSMesgQueue *D_800EA3A8[4];
-extern s32 D_800EA4A4;
+extern OSMesgQueue* D_800EA3AC[];
 
-void func_800CBC24(void)
+extern static s32 D_800EA4A4;
```
Tmcg2, line 303: This might be causing build issues. `D_800EA4A4` should be just at the top of the file as a plain `s32`, this will be read as a redefinition and the compiler is not going to like that

Tmcg2 (approved):

MegaMech (approved):

PR #375, opened 2023-08-19
Versus mode Bomb Kart stuff

Tmcg2:


data/data_0DD0A0_1.s
```diff
@@ -120,146 +120,187 @@ glabel D_800DCBB4
 .endm
 
 glabel D_800DCC08
+# Mario Raceway Bomb Kart spawns
```
MegaMech, line 123: Why did they put this data here? This should be in the course_data files.

MegaMech (approved):

data/data_0DD0A0_1.s
Tmcg2: Who knows. Maybe the bomb karts were a last minute addition? Could explain why there's an unused bomb kart portrait floating around in the data

PR #381, opened 2023-08-24
Make some adjustments to the SaveData struct

Tmcg2:
Also get rid of some fake variables from a bss file.

Matched a couple functions too.


Tmcg2:
WARNING!

The struct name "Stuff" is absolutely terrible, but its 1AM and I can't think of anything better. Any alternatives would be nice.

Member names "main" and "backup" aren't nearly as bad but are still pretty lacking. Alternatives welcome there too.

MegaMech (approved):
It's 1am and I can't think of any names.

MegaMech (dismissed):

Tmcg2:
Updated the name of `D_8018EB90` to `gSaveData`

MegaMech (approved):

PR #384, opened 2023-08-26
Some matches in in code_80086E70

Tmcg2:
I really, really don't like a couple of the matches because of
 the handling of the `unk_80` member of the object struct. But
 if it matches it matches.


Tmcg2:
I know someone else is working on this file so if they put a PR up merged theirs first I'll rebase. I don't want to step on anyone's toes.

MegaMech (approved):
lgtm!

PR #387, opened 2023-08-26
Put `void` in several function prototypes

Tmcg2:
I have historically made prototypes for argument-less functions
 with just plain `()` for the argument list. This has caused some
 minor pains when using `m2c` and is, pendatically, incorrect.

So I have changed all argument-less function protoypes to instead
 have `(void)` as the argument list. This required a couple small
 fixes here and there.


MegaMech (approved):
I think the (void) explicitly means it does not take any arguments whereas `()` means it can take any number of arguments. The compiler might output differing regalloc sometimes due to this.

Tmcg2:
I think what it is is that `()` is only a `declaration` while `(void)` is a proper `prototype`. The difference being that you can pass any number of arguments to a function that only has a `declaration` while a function with a `prototype` has a set number of arguments it'll accept.

You're right that under some circumstances the difference will cause minor differences, but the primary motivation was to get `m2c` to understand that some functions take 0 and only 0 arguments. Its not a huge pain, but its enough of a pain to make me fix it.

PR #393, opened 2023-09-01
Move ro/data for code_80091750 into the appropriate c/asm files

Tmcg2:


src/code_80091750.c
```diff
@@ -32,6 +32,1271 @@ s32 code_80091750_pad_2[2];
...
+    0x000f, 0x000f, 0x000f, 0x000f,
+};
+
+char *D_800E7500[] = {
+    "mushroom cup",
+    "flower cup",
+    "star cup",
+    "special cup",
+    "battle",
+    // ????
```
MegaMech, line 324: Might this be two or three different arrays?

MegaMech (approved):

PR #397, opened 2023-09-03
Got a couple matches and a few non matchings

Tmcg2:
Also got rid of a couple asm files for functions that were matched
 a long, long time ago (thank you ItzSwirlz for finding them)


Tmcg2:
Most (if not all) the non-matching functions in this PR are from register allocation nonsense. Given the similarity of all those functions its likely the same problem in each function. So if someone can figures out how to fix one of them they'll likely be able to fix all of them.

MegaMech (approved):

PR #407, opened 2023-09-08
Cleanup undefined_syms for all the course data

Tmcg2:


src/actors.c
```diff
@@ -20,6 +20,26 @@
...
+#include "courses/flower_cup/mario_raceway/course_data.inc.h"
+#include "courses/flower_cup/toads_turnpike/course_data.inc.h"
+#include "courses/star_cup/bowsers_castle/course_data.inc.h"
+#include "courses/star_cup/royal_raceway/course_data.inc.h"
+#include "courses/star_cup/sherbet_land/course_data.inc.h"
+#include "courses/star_cup/wario_stadium/course_data.inc.h"
+#include "courses/special_cup/banshee_boardwalk/course_data.inc.h"
+#include "courses/special_cup/dks_jungle_parkway/course_data.inc.h"
+#include "courses/special_cup/rainbow_road/course_data.inc.h"
+#include "courses/special_cup/yoshi_valley/course_data.inc.h"
```
MegaMech, line 42: I'm thinking we should have a course_data.h that contains the above. So you don't have to type every single course like this. (that can be in a future PR if someone decides to do so, and if not, that's fine too).

I don't see a reason for only including like two courses versus just always including all of them. I suppose naming conflicts could happen but I think that would be unlikely given the naming scheme. Plus probably better to catch those sooner rather than later.

MegaMech (changes requested):

src/actors.c
Tmcg2: I did consider making a top level `all_course_data.h` or something to that effect at the top level of the `courses` directory. I can do that if you want.

MegaMech: I think the all_course_data.h would be a great idea. Feel free to leave that for a future PR if you wish.

Tmcg2: I went ahead and did it. Its a small thing anyway.

MegaMech:
Whoopsee! You accidentally pushed changes to the permuter.

Tmcg2:
Fixed

MegaMech (approved):

PR #408, opened 2023-09-09
F3DEX old commands changes

AloXado320:
Due to differences between F3DEX 0.95 and 1.21, some commands have to be changed due to them still using the old F3D format used in SM64.

- `gsSPQuadrangle` uses `G_QUAD` as opposed to `G_TRI2` in later versions, this command is only used in the startup 3D logo.
- `gsSPQuadrangle` commands not in startup logo were replaced by `gsSP2Triangles` due to the above change.
- `gSPPerspNormalize` is now properly defined and works like in SM64.
- `gsSPCullDisplayList` uses the old F3D format.
- `gsDPNoOp` was replaced to `gsDPSetBlendMask`. Some N64 SDK demos use this command to setup RDP but since this command was never implemented, it gets nop'd out. It's possible that were used during `Super Mario Kart R` stages.

Some extra changes:
- Makefile: Renamed `f3d_old` to `f3dex_old`, add `f3dex` and some comments.
- `create_gfx_task_structure`: Named flags, clarify F3DLX usage and properly set `OS_YIELD_DATA_SIZE`.


include/PR/gbi.h
```diff
@@ -2161,6 +2167,26 @@ typedef union {
 	__gsSPLine3D_w1f(v0, v1, wd, flag)				\
 }}
 
+// Early F3DEX revisions use the actual quadrangle command
+// without a flag argument, this was added for compatibility.
```
MegaMech, line 2171: Could you please update this comment to be more specific. Perhaps along the lines of
```c
// mk64's startup logo uses an early revision of the quadrangle command
// It does not have a flag argument.
// Added for compatibility with F3DEX 0.95
```
The rest of the game uses a quad or tri2 which is equivallent to the newer f3dex versions.

MegaMech (changes requested):

include/PR/gbi.h
AloXado320: Both G_QUAD and G_TRI2 exist on `F3DEX 0.95`, while on 1.21 they both use `G_TRI2`. While it's possible they could have used quad on other display list, using tri2 makes them more consistent with already existing tri2 commands.

MegaMech (approved):

PR #411, opened 2023-09-09
Add new utility to calculate packed dl end and vertex count

Tmcg2:
Requires changing a bunch of actual references to packed dl lists
 with raw numbers. Which feels extremely bad but it doesn't match
 otherwise.


Tmcg2:
Doh, forgot to specify somewhere that this depends on this Python library:
https://github.com/eliben/pyelftools

So this isn't going to build right until that gets fixed. Which of course means this PR adds another requirement to the project. Don't know if that's desirable or not.

Tmcg2:
I have these packages installed in my virtualenv for the project:
```
cffi==1.15.1
colorama==0.4.6
cxxfilt==0.3.0
Levenshtein==0.21.1
pycparser==2.21
pyelftools==0.30
PyNaCl==1.5.0
python-Levenshtein==0.21.1
rapidfuzz==3.1.1
toml==0.10.2
watchdog==3.0.0
```
Most of those are for other tools (the differ and the permuter mainly). Would there be value in making a Pipfile for the project?

courses/courseTable.inc.c
```diff
@@ -13,152 +13,145 @@
  * This tool is ran during the compile process.
 */
```
MegaMech, line 14: Is this comment still applicable?

MegaMech (changes requested):
Can you update the readme with the dependency?

Tmcg2:
I actually originally wanted to use something like `readelf` or `objdump` or `size` but all of them share the same problem. They only really allow you to get the size of the data section including the padding at the end. Which is unfortunate because we can't predict how much padding there is, meaning we run the risk of off-by-one errors (see the vertex count utility).

We might be able to avoid a new dependency if I rewrote them in C. I presume (but do not know for sure) that the necessary header file (elf.h) is probably distributed with other things in the GNU utils. But I'm lazy and enjoy having Python handle all the file nonsense. I also assume this Python library is just python bindings around elf.h with some extra sugar here and there.

MegaMech:
Can't we just link these variables now and use the array count macro and for the final displaylist just add its size to itself?

Tmcg2:
I think so, but it would be a little weird.

If in a header file you have an array like `extern Gfx some_array[];` you can't actually do `sizeof(some_array)` in a different file. The fact that the array has no definite size (in the header file) means its an incomplete type. In order for the `ARRAY_COUNT` macro to work the array would need to have its size specified, something like `extern Gfx some_array[0x17C0]`.

So, we could make a `model.inc.h` file for each course (or maybe just an `all_models.inc.h`?) that has the arrays with a definite size. But then we're no longer programmatically detecting those sizes, so any adjustments to those arrays will cause things to break. That's not a huge problem, especially for a decomp where those things don't change, but it'll be a hassle for anyone looking to make changes down the road.

Similar story with the packed_dl stuff. We'd have to have to give a definite size to the very  last DL list (at minimum) in the header file and then do something like `((uintptr_t) my_dl_array) + ARRAY_COUNT(my_dl_array) -  8 0x07000000`. But again, that means any changes to that last DL list will likely cause things to break.

None of this is to say that doing any of the above things are bad. They're just "brittle", for lack of a better term.

MegaMech:
Would C's type checking prevent changes to these arrays without updating the header size?

(Does C allow you to lie and put in a wrong size?)

Tmcg2:
If we don't include the `model.inc.h` file in the `model.inc.c` file we can probably lie about the array's size all we want. I'd have to test it to be sure but that sounds right to me.

And  yeah, if we're okay with some brittleness we can just leave it as a comment in the header file(s) warning people about the potential pitfalls of modifying those arrays.

Tmcg2:
Removed the old and new generator utilities, hardcoded the array lengths in the appropriate header files. If those numbers stop being accurate stuff will break.

Tmcg2:
Sorry for all the force pushes, should be done now

MegaMech (approved):

PR #413, opened 2023-09-10
Assorted matches in audio/external

Tmcg2:


Tmcg2:
This has some overlap with Swirlz's PR, so one of is going to have to rebase. I'm pretty comfortable with doing that if it makes things easier.

MegaMech (approved):

MegaMech (approved):

PR #422, opened 2023-09-16
Add asset type for (mk64_)vtx

Tmcg2:
That way we can extract things like the course models and other
 vertex lists directly from the rom instead of hard-coding them
 in various data files


Tmcg2:
Adding this as a draft for now to see if there's an actual appetite for this. I don't think it really matters if we hard-code all the vertex lists.

MegaMech:
Took another look at this PR.

I think it's in the right direction. But we should just generate the course_data file in its entirety. And in cpp for future compatibility.

PR #426, opened 2023-09-18
Moved and renamed course files

MegaMech:
This PR deletes the cup folders. This game is simple enough it's not necessary, the current approach over-complicates the build system and makes it less mobile for the future.

Renamings:
model.inc.c --> vtx.inc.c
packed.inc.c --> displaylists.inc.c (or triangles or material.inc.c?)

packed name will only be used *after* the displaylists are packed. Probably mtl.packed.bin or something.

course_data name will stay the same.

This is breaking the build. It runs before the vtx mio0 rule can run.
Not sure on a solution. But haven't had any time to think about one.
```makefile
$(BUILD_DIR)/%.mio0.s: $(BUILD_DIR)/%.mio0
	$(call print,Generating mio0 asm:,$<,$@)
	printf ".section .data\n\n.balign 4\n\n.incbin \"$<\"\n" > $@
```


MegaMech:
Closed cause abandoned.

PR #436, opened 2023-09-21
cleaning camera, kart_dma, menu, math_util

coco875:


src/menus.h
```diff
@@ -11,6 +11,13 @@ enum MenuOptionsCursorPositions {
     MENU_OPTIONS_CSRPOS_SOUNDMODE = 0x16
 };
 
+// needs to be a union (or array...?) to go into rodata as a const
+// terrible for endianness... Best guess as to what this is for..
+union GameModePack {
+    u8  modes[4];
+    s32 word;
+};
```
MegaMech, line 19: If GameModePack is only used in menus.c the convention would be to have it in the .c

If it's used in other files, then the convention would be to have it in .h.

MegaMech (changes requested):

src/menus.h
coco875: it's also used in the header file for D_800F2BE4

MegaMech (approved):

PR #440, opened 2023-09-22
Function renaming for math_utils and math_utils_2

coco875:


src/math_util_2.c
```diff
@@ -342,6 +342,7 @@ Vec3f *vec3f_set_dupe_2_electric_boogaloo(Vec3f arg0, f32 arg1, f32 arg2, f32 ar
     return (Vec3f *) &arg0;
 }
 
+// look like normalise vector
 Vec3f *func_80041530(Vec3f dest) {
```
MegaMech, line 346: Please name this func `normalize_vector` (please check there isn't already a func named `normalize_vector`)

// @warning Returns addr of local variable.

src/math_util_2.c
```diff
@@ -478,117 +479,124 @@ UNUSED void func_80041A70(void) {
...
-    arg0[3][3] = 1.0f;
+    /*
+     * 1 0 0 x
+     * 0 1 0 y
+     * 0 0 1 0
+     * 0 0 0 1    
+     */
+}
+
+void mtxf_u16_rotate_z(Mat4 dest, u16 angle) {
```
MegaMech, line 508: I like the name. I believe angles in n64 are always u16.

Just wondering if `mtxf_rotate_z_from_angle` might be better? But your name is also good.

src/math_util_2.c
```diff
@@ -478,117 +479,124 @@ UNUSED void func_80041A70(void) {
...
+    dest[0][3] = 0.0f;
+    dest[1][3] = 0.0f;
+    dest[2][3] = 0.0f;
+    dest[2][2] = 1.0f;
+    dest[3][3] = 1.0f;
+    dest[0][0] = scale;
+    dest[1][1] = scale;
+}
+
+void mtxf_rotate_z_scale_x_y(Mat4 dest, u16 angle, f32 scale) {
```
MegaMech, line 549: Could you please add `UNUSED` to this func?

`UNUSED void mtxf...`

(yea, unfortunately it's called by an unused func)

src/racing/math_util.c
```diff
@@ -161,34 +165,44 @@ void mtxf_identity(Mat4 mtx) {
...
+     * mat(2,0)        mat(2,1)        mat(2,2)        mat(2,3)
+     * mat(3,0)+pos(0) mat(3,1)+pos(1) mat(3,2)+pos(2) mat(3,3)
+     */
 }
 
-UNUSED void func_802B54EC(Mat4 arg0, Mat4 arg1, Vec3f arg2) {
-    arg1[3][0] = arg0[3][0] + arg2[0];
-    arg1[3][1] = arg0[3][1] + arg2[1];
-    arg1[3][2] = arg0[3][2] + arg2[2];
+// more lite version of add_translate_mat4_vec3f
```
MegaMech, line 198: Light version...

(remove more)

src/racing/math_util.c
```diff
@@ -426,18 +464,20 @@ void func_802B5D64(uintptr_t arg0, s16 arg1, s16 arg2, s32 arg3) {
...
-    arg0[0][0] *= arg1;
-    arg0[1][0] *= arg1;
-    arg0[2][0] *= arg1;
-    arg0[0][1] *= arg1;
-    arg0[1][1] *= arg1;
-    arg0[2][1] *= arg1;
-    arg0[0][2] *= arg1;
-    arg0[1][2] *= arg1;
-    arg0[2][2] *= arg1;
+// multiply a matrix with a number
```
MegaMech, line 467: // Scale 3x3 of a 4x4 matrix

src/racing/math_util.c
```diff
@@ -426,18 +464,20 @@ void func_802B5D64(uintptr_t arg0, s16 arg1, s16 arg2, s32 arg3) {
...
+void mtxf_multiplication_number(Mat4 mat, f32 coef) {
```
MegaMech, line 468: `mtxf_scale_3x3` ?

MegaMech (changes requested):

src/math_util_2.c
coco875: you have also func_802B6374 who do the same thing

src/math_util_2.c
coco875: I name u16 because you have a function who use instead s16 so to differenciate between this two function I add u16/s16

src/racing/math_util.c
coco875: but he use a mat4 and in general we don't touch the column four and the line four when you need to multiply with a position (vec3)

MegaMech: What about scale rather than multiply? Or do you think multiply is more accurate?

src/math_util_2.c
MegaMech: Okay, this is fine then.

MegaMech (approved):

src/math_util_2.c
MegaMech: good good

src/racing/math_util.c
coco875: by some research it's appear to be scale

MegaMech (approved):

PR #443, opened 2023-09-27
Proper linkage for textures in course display lists

Tmcg2:
Or, at least, for Banshee Boardwalk

There will have to be files like this for every single course.

I have elected to not move the extraction of these textures to
 new_extract assets in order to keep the scale of these chagnes
 to a minimum. Its not necessary to do so anyway, although it
 would make me feel better.

These changes will also likely break @coco875 's course extraction
 stuff, which might be a good enough reason to decline such changes


coco875:
It can be easely fix because I already catch other texture in C for other entity

Tmcg2:
Rereading my comment, I should clarify: "decline such changes" is in reference to this PR, not coco's stuff.

MegaMech:
So, we can actually split the other_textures.s file into its relevant courses?

Tmcg2:
No, we can't get all of other_textures matched like this. It is absolutely a hack but its the only way to get stuff linked properly. We can of course just not care about linkage, things work fine without, this is just here to show what a fix for that would have to look like.

To try and summarize what's going on:
Each course has a `course_name-here_offsets` file (I think you made them?) and each of those has a `course_texture` array. The games steps through that array whenever a given course is loaded. For each entry, the texture is decompressed into Segment 5. Segment 5 then is unique for each course. There can't be a 1 file solution because different courses will have different textures at the same address.

For example:
```const course_texture banshee_boardwalk_textures[] = {
    {gTexture6447C4, 0x0106, 0x0800, 0x0},
    ...
const course_texture choco_mountain_textures[] = {
    {gTexture64619C, 0x0124, 0x0800, 0x0},
```
Choco Mountain and Banshee Boardwalk have a different texture at address `0x05000000`. So they  can't use the same C file for linkage purposes. And of course, even if two courses use the same texture somewhere, they may be at different addresses in each course.

MegaMech:
Ah, I see. This is likely the _only_ solution if we want linkage.

Which direction are you leaning towards?

One possible downside if people modify the textures file and forget to modify the linked file.

I think this is a great example of the solution. But I'm not sure if we should go all the way until we're sure it's not going to be replaced with a different solution in the near future.

Tmcg2:
Its the only solution I can think of.

I don't have too strong of an opinion on the matter. We clearly don't NEED linkage for things to work and really any future port/modding can probably change things enough to handle this in a cleaner manner (or at least, I'd hope they can).

If someone made a change to the course_offsets stuff they'd probably have to change things in the display lists anyway so I'm don't think it'd be too big a hassle to have them also change stuff in the linkonly file.

MegaMech:
So, thinking about this further.

Both solutions, #defining and linking result in the same issue; a file that would need updating.

Some fancy tool could automate this, but thing with tools like this is they can break. This is clearly a better solution then #defining the variables.

Tmcg2:
I really wish there was a better way to get the dependency lists for each course. In theory a `.d` file should handle that, but I don't get how to make the `inc.c` files generate from a clean build state. They have to exist before the `.d` file and I can't figure out how to make that happen. Its probably user error on my part, the way I have stuff now works, but its fugly.

Tmcg2:
What's Jenkins complaining about? Works on my machine so there might be something I forgot to add to the PR laying around in my build tree

Tmcg2:
Took a cue from @AloXado320's PR and added an auto-generator for the linkonly files. That way we have the best of both worlds: Alo's autogeneration with proper linkage.

The downside is that there's that really nasty texture-to-filename map in the generator itself. But I don't see a way around it.

MegaMech:
I haven't looked at this PR for a bit.

Should I take another look?

MegaMech (approved):

PR #444, opened 2023-09-28
Doc setup_game_memory, label syms, add osSyncPrintf support, match funcs

MegaMech:
Adds is_debug.c which enables support for osSyncPrintf
under the DEBUG flag. Should enable with make DEBUG=1

Not the best solution for now. Because that probably requires make clean first.

Renames:
```
segments.h        --> hardcoded_segments.h
segment_symbols.h --> segments.h
```
hardcoded_segments.h is only for the linker.

segments.h has avoid_ub ifdef to replace hard-coded addr with proper linker symbols.


ItzSwirlz:
Might be a good idea to setup a GitHub Pages website for this

PR #447, opened 2023-10-05
Match func_80026B4C and fix D_802F1F80

AloXado320:
Matches the last file remaining on `code_8001F980` courtesy of an anonymous user on decomp.me

Also fixes `D_802F1F80` struct variable and code pointers under the `AVOID_UB` define.

Note that there's remaining struct variables to be fixed but this should be as a starting point for that.


src/code_8001F980.c
```diff
@@ -1372,7 +1372,11 @@ void func_800248D0(Player *player, s8 arg1, s8 arg2, s8 arg3) {
             sp154[2] = player->pos[2] + sp140;
         }
     }
-    D_80164B04 = &D_802F1F80[D_801651D0[arg2][arg1]][arg2][arg1 << 7];
+#ifdef AVOID_UB
+    D_80164B04 = &D_802F1F80[D_801651D0[arg2][arg1]][arg2][arg1].kart_palette;
```
Tmcg2, line 1376: Not that this really matters, but I don't think this function is loading just the `kart_palette`. I think it wants the address of the entire `D_802F1F80` entry given its use in the `gDPLoadTLUT_pal256` call. Granted those two things will have the same address in practice, so this is mostly a pedantic nitpick.

AloXado320: I though the name "kart" was in general since it was just there, but i guess it would make sense if both kart and wheel use the same tmem memory, i still find it weird how they use a different offset for the wheel pallete though.

Tmcg2: Oh, there's an entire thing going on here. I will try to summarize it (my summaries are usually bad, so feel free to ask for clarifications)

Animation frames for the characters are CI8 (palette using) images. But the palettes used by those images are unique. Instead of a palette per frame, there's a so-called "kart palette" that takes up the upper 3/4's of the palettes color space. The bottom 1/4 is reserved for the "wheel palette". There is an obvious misnomer here: aren't the wheels part of the kart? In this explanation, no.

The entries in `D_802F1F80` are full palettes, but they have component parts, the kart and wheel parts. The way the game animates the wheels (to give the illusion of them rolling) is to rapidly cycle the wheel part of the palette through a set of 1/4 sized palettes (4 per animation frame). That's what `func_80026B4C` is responsible for, choosing which "wheel palette" to use.

MegaMech: What name would you suggest instead of `kart_palette`?

src/code_8001F980.c
```diff
@@ -1618,61 +1626,57 @@ void func_80026A48(Player *player, s8 arg1) {
...
-#ifdef MIPS_TO_C
-//generated by m2c commit b52d92c2340f6f4ba1aafb464188bb698752fbb0 on Jul-30-2023
-typedef u8 wheels[4][128];
-
-extern wheels gKartMario168Wheel0;
-
-extern wheels *gKartMarioWheels0[];
+// Properly define struct pointers, see framebuffers.h comment for more information.
+#ifdef AVOID_UB
+#define D_802F1F80_WHEEL(a, b, c) &D_802F1F80[a][b][c].wheel_palette
```
MegaMech, line 1631: Shouldn't struct member names be `camelCase` not `snake_case`? I don't mind an exception here if there's a good reason for it though.

MegaMech (changes requested):

src/code_8001F980.c
Tmcg2: The name `kart_palette` is fine, what I was originally getting at was that `&D_802F1F80[D_801651D0[arg2][arg1]][arg2][arg1].kart_palette` would be better replaced with `&D_802F1F80[D_801651D0[arg2][arg1]][arg2][arg1]` since that's """more correct""". But, as originally mentioned, that's a pedantic nitpick, its doesn't actually matter. It just bothers my aesthetic sensibilities.

src/code_8001F980.c
Tmcg2: No reason, just what I came up with when I originally added the struct.

Tmcg2:
Yeah that looks fine. 👍

src/framebuffers.h
```diff
@@ -39,7 +39,7 @@ typedef struct {
  *   kart_palette appears to be the palette for all non-wheel elements of the kart (including the driver).
  *   wheel_palette then is the palette for the wheels
  *
- * kart_palette does not appear to be used 100% of the time, more testing is needed to figure out what's going on
+ * kart_palette is "used" in the sense that calling the struct variable directly uses it
```
MegaMech, line 42: This seems vague and feels like a response to the line you deleted.

I'm not sure if it makes sense by itself.
`* The calling struct variable directly uses the kart_palette` Maybe something along these lines. But this still doesn't make sense to me.

Perhaps something like
`Once loaded, the kart palette is used for the entire race whereas a variety of wheel palettes are chosen as you drive around.`

Perhaps this line is already described below and the line in-question could be deleted?

MegaMech (changes requested):

src/framebuffers.c
```diff
@@ -13,7 +13,11 @@ u8 frameBufferPadding[22544];
 
 struct_D_802BFB80 D_802BFB80[2][2][8];
 struct_D_802DFB80 D_802DFB80[2][2][8];
-u32 D_802F1F80[2][4][1024];
+#ifdef AVOID_UB
+struct_D_802F1F80 D_802F1F80[2][4][8];
```
MegaMech, line 17: If you wanted to take this one step farther, you could name the struct and variable. Your choice.

MegaMech (changes requested):

src/framebuffers.c
AloXado320: That would be better for a future PR to document the file

src/framebuffers.h
AloXado320: I tried to say that while `kart_palette` is not "called" directly, it is used when you call the struct itself, recall what Tmcg2 said regarding it's usage, i can edit the comment to be more clear regarding that.

MegaMech: I read the *whole* comment.

Not super clear. But does okay. Someone can improve it later if they want.

MegaMech (dismissed):

src/framebuffers.h
```diff
@@ -36,11 +36,11 @@ typedef struct {
...
- *   kart_palette appears to be the palette for all non-wheel elements of the kart (including the driver).
- *   wheel_palette then is the palette for the wheels
- *
- * kart_palette does not appear to be used 100% of the time, more testing is needed to figure out what's going on
- * wheel_palette changes as you drive around while kart_palette is constant
+ *   kart_palette contains the pallete for all non-wheel elements of the kart (including the driver).
+ *   wheel_palette contains the palette for the wheels.
+ * 
+ * kart palette sets a defined palette based on the character.
+ * while wheels pallete sets a dynamic palette as your drive around with the kart.
```
MegaMech, line 43: ~~your~~ you
`wheel_palette sets a dynamic palette as you drive around with the kart.`

MegaMech (changes requested):

MegaMech (approved):

MegaMech (approved):

PR #448, opened 2023-10-07
Rename wayPoint to waypoint

### LINKED ISSUE #442: WayPoint needs to be renamed
Waypoint should be considered a single word imo.

Therefore all wayPoint should be renamed to waypoint. Also wayPointX --> waypointX or even just posX 

TrackWayPoint --> TrackWaypoint

MegaMech:
Also typedef TrackWaypoint struct and rename it as well

Resolves #442


src/racing/render_courses.c
```diff
@@ -113,71 +113,87 @@ void load_surface_map(uintptr_t addr, struct UnkStruct_800DC5EC *arg1) {
     }
     arg1->playerDirection = var_a3;
 
-    if (D_80152300[temp_a2 - camera1] == 1) {
```
Tmcg2, line 116: Is this deletion meant to be part of the PR?

MegaMech: Naa, I added an enhancement I need to remove.

PR #449, opened 2023-10-08
Generate header defines for segment 5 textures

AloXado320:
A simplify take on #443

It relies on a python script `mkseg5defs.py` to generate a header by using a struct texture table in `course_offsets.c`, it uses a variable set in the table as the decompressed size to calculate the offset of the texture since segment 5 is generated by the game.

I also did other changes such as moving the directory of the offset c files so they are in the `courses` folder, and removing some unnecessary headers in display list course files.


tools/mkseg5defs.py
```diff
@@ -0,0 +1,60 @@
+import os
+import sys
+
+# This script converts MK64 course_texture struct tables to segment 5 address defines.
```
MegaMech, line 4:
```
# This script generates course/<course_name>/course_offsets.h files from mk64 course_texture struct arrays. The headers contain segment five defines.
# Segmented five contains the textures required to render the course. They are selected from other_textures.s using the array in course_offsets.inc.c. Each course has its own selection of textures to use.
```

tools/mkseg5defs.py
```diff
@@ -0,0 +1,60 @@
...
+# Segment 5 addresses are generated by the game using a texture course table.
+# This uses data size (uncompressed size) value on the same table to generate the address.
+
```
MegaMech, line 7: Please add how to use the script

`# python3 mkseg5defs.py <offsets_input_file> <header_output_file>`

MegaMech (changes requested):
Not a fan of the python script name mkseg5defs nor makefile variable of the same name.

generateCourseOffsetDefinitions.py ? Or something along those lines?
makeCourseOffsetsDefinitions.py?


Also, port to c++ when?

tools/mkseg5defs.py
AloXado320: I was documenting how the tool works and i forgot to add a "help" argument lol

AloXado320:
> Not a fan of the python script name mkseg5defs nor makefile variable of the same name.
> 
> generateCourseOffsetDefinitions.py ? Or something along those lines? makeCourseOffsetsDefinitions.py?
> 
> Also, port to c++ when?

I went from a shorter "simple" like the rest of the tools, also it only takes the texture table from course offset file.

If you want to get verbose it could call the tool `generateCourseTextureAddress.py` lol

MegaMech:
It doesn't need to be that verbose. I don't like the seg5 part.

The other tools use snake_case.

AloXado320:
After a bit of research, it would be better to generate the whole texture course table along with their addresses, so i'm closing this.

Issue #455, opened 2023-10-11
Incomplete Doxygen documentation


coco875:
When generate doxygen documentation not all file and function appears

coco875:
a issue with https://github.com/n64decomp/mk64/pull/458 it's he document all function two time because one are in .h and another in .c because of EXTRACT_ALL in doxygen file but if we don't enable it all function undocument will not appears to fix this we need to document all function and set EXTRACT_ALL to NO when finish

PR #467, opened 2023-10-15
Textures doc

coco875:
add comment in texture for https://github.com/n64decomp/mk64/issues/455


src/data/textures.inc.c
```diff
@@ -1,671 +1,1212 @@
...
+ * look like link to code_80091750
+ */
 #include "textures.h"
 
-// 0x801978D0
+/**
+ * @brief MkTexture of a tiny font texture of number 0
+ *
+ */
 MkTexture D_02000000[2] = {
```
MegaMech, line 13: Wonder if perhaps these don't need comments and just names.

But merge cause it's fine like this too.

MegaMech (approved):

src/data/textures.inc.c
coco875: when the extrat_all are not enable he all with comment or this include .h file but he are needed to have struct so we need to comment all variable and function to disable extract_all and have less redundencies

PR #470, opened 2023-10-16
Match audio/external.c BSS segment

Tmcg2:
Also fix a misnamed function found in audio/heap, used in
 audio/load


Tmcg2:
Barring having to fix a build issue, I'm done touching this PR.

Tmcg2:
And there's no build issue. Excellent.

src/audio/seqplayer.c
```diff
@@ -10,6 +10,41 @@
...
+char seqplayer_unused_string17[] = "SUB:ERR:BANK %d NOT CACHED.\n";
+char seqplayer_unused_string18[] = "Audio:Track: CTBLCALL Macro Level Over Error!\n";
+char seqplayer_unused_string19[] = "[%2x] \n";
+char seqplayer_unused_string20[] = "Err :Sub %x ,address %x:Undefined SubTrack Function %x";
+char seqplayer_unused_string21[] = "Disappear Sequence or Bank %d\n";
+char seqplayer_unused_string22[] = "Macro Level Over Error!\n";
+char seqplayer_unused_string23[] = "Macro Level Over Error!\n";
+char seqplayer_unused_string24[] = "Group:Undefine upper C0h command (%x)\n";
+char seqplayer_unused_string25[] = "Group:Undefined Command\n";
+
```
MegaMech, line 47: So, is it correct that, some of these were placed in _reasonable_ spots. But you imported more of them so now they all go up here cause effort to figure out where they all go?

I'm fine with that. Just checking.

MegaMech (approved):

src/audio/seqplayer.c
Tmcg2: Yes. The placement of the stubbed print's is almost certainly correct (probably just copied from SM64) but trying to get ALL the strings in the right place while respecting those placements would be a lot of effort. So I opted for the slightly lazy option of commenting them out and moving all the strings to the top of the file.

PR #478, opened 2023-10-25
Linkage for the ceremony_data file

Tmcg2:
Cleaned up all but 1 undefined symbol from the 0x0B range. The one
 leftover may be something from texture_tkmk00. Or atleast, that's
 what I think. Can't say for certain


Tmcg2:
Cleaned up a lot of undefined symbols related to other_textures as well.

MegaMech (approved):

MegaMech:
Does your PR delete `coursePath` I'm presuming it's now unused

MegaMech (approved):

PR #479, opened 2023-10-28
match func_8000F628 and add NUM_PLAYERS define

MegaMech:


Tmcg2:
Think you forgot to delete that asm file for the function

PR #487, opened 2023-11-03
Matches in code_80005D0

MegaMech:


Tmcg2:
My only request would be to add the newly matched functions and any `extern`d variables they require to the appropriate header. That way any future uses of mips_to_c will have them as context which may help make its output easier to understand

MegaMech:
Will do

PR #488, opened 2023-11-04
More audio stuff

Tmcg2:
Due to the hack job I've done to the `abi.h` file, I'm convinced that we don't need nor want the EU/Shindou/US ifdefs in that file. I suspect all version of MK64 use the same `abi` stuff and we need to figure out what all that looks like. Unfortunately I don't know how to go around figuring that out except by running into it when decomping audio functions.


MegaMech:
There's an unifdef utility where you can strip out all the ifdefs you don't want.

MegaMech (approved):

MegaMech (approved):

PR #491, opened 2023-11-08
Upgrade Dockerfile to Ubuntu 22.04 as 18.04 is EOL

MisterSheeple:
In my testing, there appears to be zero breakage. It successfully builds a matching ROM without any further adjustments other than bumping the Ubuntu version. However I'm being told some parts of the build chain are outdated so I'm going to update those as well.

- [x] Use recomp instead of qemu-irix IDO compiler
- [x] Bump dependencies to use current ones from the README
- [x] Add documentation to the README


MegaMech (approved):

MisterSheeple:
You guys kind of have a messy readme here so I tried my best, but let me know if what I wrote should be changed.

Also, if someone could please re-review my pull with this new information in mind, I'd really appreciate it.

MegaMech (approved):

PR #495, opened 2023-11-10
Import some code_80071F00 matches

ItzSwirlz:
Credit goes to Taggerung


Tmcg2:
My only request would be to put the prototypes for the matched functions and any `extern`d variables in the appropriate header file. That way they can be picked up as `mips_to_c` context which helps makes its output more accurate.

src/code_80071F00.c
```diff
@@ -8342,43 +8333,30 @@ void func_80086528(s32 objectIndex, s32 arg1) {
     }
 }
 
-#ifdef MIPS_TO_C
-//generated by m2c commit eefca95b040d7ee0c617bc58f9ac6cd1cf7bce87 on Aug-10-2023
 extern s16 D_8016347C;
 extern s8 D_801658C6;
```
MegaMech, line 8337: Please delete these two externs now that they are in the header.

src/code_80071F00.c
```diff
@@ -5247,26 +5247,17 @@ void func_8007E00C(s32 objectIndex) {
 GLOBAL_ASM("asm/non_matchings/code_80071F00/func_8007E00C.s")
 #endif
 
-#ifdef MIPS_TO_C
-//generated by mips_to_c commit 3c3b0cede1a99430bfd3edf8d385802b94f91307
-void func_8007DDC0(s32); // extern
-void func_8007E00C(s32); // extern
 extern s32 D_80183EA4;
```
MegaMech, line 5250: Please delete this extern now that it's in the header.

MegaMech (changes requested):

MegaMech (approved):

PR #506, opened 2023-12-24
fix doxygen syms

coco875:
change to be a md file and display properly the table
image


MegaMech:
Looks great!

I might see if we can get rid of the zeros so it's just 0x8 instead of 0x000000008

MegaMech (approved):

PR #510, opened 2023-12-24
Doxygen Syms Gen Ignores Switch Jumptables

coco875:


tools/doxygen_symbol_gen.py
```diff
@@ -27,6 +29,9 @@ def process_map_file(map_file_path):
 
                     function_name = tokens[-1]
 
+                    if re.match(r"L[0-9a-zA-Z]+", function_name):
```
MegaMech, line 32: This results in false positives.

Do

L8

Because no one is going to name a function L8 and I believe that should catch all the switch jumptables.

MegaMech (changes requested):

MegaMech (approved):

PR #518, opened 2023-12-30
Add bss render player and spawn player

coco875:


src/spawn_players.c
```diff
@@ -22,6 +22,50 @@
 #include "menus.h"
 #include "render_player.h"
 
+f32 D_80165210[8];
+f32 D_80165230[7];
+f32 D_8016524C;
```
MegaMech, line 27: It's possible that D_8016524C is apart of the D_80165230 array

src/spawn_players.c
```diff
@@ -22,6 +22,50 @@
...
+bool gIsPlayerTripleAButtonCombo[8];
+s32 D_80165480[8];
+s32 D_801654A0[8];
+s32 D_801654C0[8];
+s32 D_801654E0[8];
+s32 D_80165500[8];
+s32 D_80165520[8];
+s32 D_80165540[8];
+s16 D_80165560[7];
+s16 D_8016556E;
```
MegaMech, line 57: It's possible that D_8016556E is apart of the D_80165560 array

MegaMech: Took a look at the code. I think your PR is correct

src/spawn_players.c
MegaMech: I couldn't get either direction to match. This is a weird contradictory variable duplication and array overflow. D_80165230 should technically be size of 8 not 7.

MegaMech (approved):

src/spawn_players.c
coco875: yeah I also see that but when I remove D_8016524C and set a size of 8 It's not match so maybe need to be fix later

src/spawn_players.c
coco875: I also try but get a missmatch so weird but add a comment to prevent this weird situation and for somone else who have more experience fix it

MegaMech (approved):

PR #524, opened 2024-01-01
some rename in player controller and spawn players

coco875:


src/code_8008C1D0.c
```diff
@@ -303,10 +303,10 @@ void func_8008C73C(Player *player, s8 arg1) {
         D_80165190[2][arg1] = 1;
         D_80165190[3][arg1] = 1;
         D_80165280[arg1] = player->currentSpeed;
-        D_80165480[arg1] = 0;
+        gTimerBoostTripleACombo[arg1] = 0;
         gIsPlayerTripleAButtonCombo[arg1] = FALSE;
-        D_80165440[arg1] = 0;
-        D_80165420[arg1] = 0;
+        gCountAChangement[arg1] = 0;
```
MegaMech, line 308: What does this mean/do (Changement is not a word)

src/player_controller.c
```diff
@@ -3428,124 +3427,124 @@ void func_80031F48(Player *player, f32 arg1) {
     }
 }
 
-void func_80032050(Player *player) {
-    s32 var_v0;
+void detect_triple_b_combo_b_released(Player *player) {
```
MegaMech, line 3430: How about:

detect_triple_b_combo_released


The idea is that the whole combo is pressed/released.

I think _b_released is somewhat repetitive. The function comment could always describe more thoroughly what's going on if the function name isn't descriptive enough.

MegaMech (changes requested):

src/player_controller.c
coco875: detect_triple_b_combo_released can let think of releasing the combo and not the button b

src/code_8008C1D0.c
coco875: oh my bad x)

MegaMech (approved):

PR #527, opened 2024-01-02
Finish bss

coco875:
all bss have been put in a file there is just code_80057C60.c who are weirdly big


coco875:
some need to be modify to be interpreted by doxygen

Tmcg2:
Could the changes to the `objects` and `bomb_kart` files be undone? Not everything needs to be put in the header file of the file they're defined in. Their placement in other header files was to try and make the includes a bit more logical. So if you want object stuff you include `objects.h` instead of having to know the specific code file the variable was defined in.

src/code_80057C60_var.c
```diff
@@ -0,0 +1,227 @@
...
+Vec3s D_8018D770[8];
+Vec3s D_8018D7A0[8];
+Vec3s D_8018D7D0[8];
+
+Vec3s D_8018D800[8];
+Vec3s D_8018D830[8];
+Vec3s D_8018D860[8];
+Vec3s D_8018D890[8];
+s16 gPlayerBalloonCount[8];
+Vec3s D_8018D8D0[8];
```
MegaMech, line 227: Please add a new empty line at the end of this file.

coco875:
so just revert change in object.h ?

Tmcg2:
And `bomb_kart.h`

coco875:
ok I will do it

src/code_8006E9C0.c
```diff
@@ -172,7 +173,7 @@ void func_8006EEE8(s32 courseId) {
 // I really really don't know what D_8018D9B4 is supposed to be
 // Its value is a pointer that seems to point to compressesed data
 // of some variety (sometimes TKMK, sometimes MIO0)
-extern u8 *D_8018D9B4;
+extern uintptr_t *D_8018D9B4;
```
MegaMech, line 176: Usually the type here should be what the data points at.

I think u8 is more accurate. But I'm not super concerned either.

MegaMech (changes requested):

MegaMech:
Great work on this PR.

This is great!

coco875:
@Tmcg2 for decleration in object and bomb_kart I also reput the comment or I just make a duplication of the comment ?

src/code_8006E9C0.c
coco875: with u8 match

src/code_80057C60.c
```diff
@@ -364,7 +366,6 @@ f32 D_8018CFCC;
 s16 D_8018CFD0;
 f32 D_8018CFD4;
 s16 D_8018CFD8;
-UNUSED s32 D_8018CFDC;
```
MegaMech: This might be why it doesn't match now.

MegaMech (changes requested):

src/code_80057C60.c
coco875: it auto align in 0x8 so a file bss can only start with XXXXX8 or XXXXX0 so it's not needed here

MegaMech (approved):

PR #529, opened 2024-01-03
some const placement and one rename

coco875:


src/actors/falling_rocks.inc.c
```diff
@@ -63,9 +63,9 @@ void update_actor_falling_rocks(struct FallingRock *rock) {
     func_802ADDC8(&rock->unk30, 10.0f, rock->pos[0], rock->pos[1], rock->pos[2]);
     pad0 = rock->unk30.unk3C[2];
     if (pad0 < 0.0f) {
-        unkVec[0] = -rock->unk30.unk60[0];
-        unkVec[1] = -rock->unk30.unk60[1];
-        unkVec[2] = -rock->unk30.unk60[2];
+        unkVec[0] = -rock->unk30.orientationVector[0];
+        unkVec[1] = -rock->unk30.orientationVector[1];
+        unkVec[2] = -rock->unk30.orientationVector[2];
```
MegaMech, line 68: rot or rotation?

src/code_80057C60.c
```diff
@@ -1,3 +1,8 @@
+/**
+ * @file code_80057C60.c
+ * @warning there is to many variable here
```
MegaMech, line 3: there are too many variables here

MegaMech (changes requested):

src/actors/falling_rocks.inc.c
coco875: it's not a rotation with u16 is more close of an orientation between 0 and 1 in float like cos and sin so an orientation normalised

MegaMech (approved):

PR #530, opened 2024-01-03
rename some texture

coco875:


src/data/common_textures.inc.c
```diff
@@ -2566,92 +2566,92 @@ f32 D_0D009808[][4] = {
 };
 
 // Odometer speedometer
-s8 D_0D009958[] = {
-    #include "textures/common/132B50_09958.i4.inc.c"
+s8 gTextureSpeedometer[] = {
```
MegaMech, line 2569: Teeeechnically Odometer... But I think most people would use speedometer so we'll stick with that.

src/data/common_textures.inc.c
```diff
@@ -34,8 +34,8 @@ s8 gTLUTFinishLineBanner[] = { // tlut
     #include "assets/finish_line_banner/gTLUTFinishLineBanner.inc.c"
 };
 
-s8 D_0D000200[] = { // tlut
-    #include "textures/common/132B50_00200.rgba16.inc.c"
+s8 gTextureParticleFire[] = { // tlut
```
MegaMech, line 37: common_texture_fire_particle

Please use this format for the common_textures.inc.c: snake_case, prefix common_

MegaMech (changes requested):

src/data/common_textures.inc.c
coco875: oh ok I forget

coco875: so it's should be the same with gTLUTFinishLineBanner no ?

MegaMech: yes

coco875: same for name of file ?

MegaMech: Just for variables in common_textures.inc.c should start with common_

The idea is that when you see a variable that starts with common_ you know that the variable contains .data. And is common for every course, so it's safe to use that variable no matter the context.

Versus if you see `d_luigi_raceway` you know that in-order to safely use this variable you need to do the following::
```c
switch(gCurrentCourse) {
    case LUIGI_RACEWAY:
}
```
Before it's safe to use that variable.

MegaMech:
Can we change _TLUT_ to _tlut_ and _HUD_ to _hud_ ?

MegaMech:
\_minimap\_ instead of \_mini\_map\_

coco875:
> Can we change _TLUT_ to _tlut_ and _HUD_ to _hud_?

I don't know like it's acronym if we make it in uppercase people will understand we abbreviate this

src/data/common_textures.inc.c
iProgramMC: The odometer measures the total distance the vehicle has traveled.

MegaMech: I'd like to retract my previous statement.

MegaMech:
Lets use lower-case for acronyms

coco875:
> Lets use lower-case for acronyms

Ok I will do this

MegaMech (approved):

src/main.h
```diff
@@ -10,12 +10,12 @@
 #define MESG_START_GFX_SPTASK 103
 #define MESG_NMI_REQUEST 104
 
-#define MTX_HUD_POOL_SIZE 800
+#define MTX_hud_POOL_SIZE 800
```
MegaMech, line 13: This should stay all caps because defines are all caps

src/main.h
```diff
@@ -10,12 +10,12 @@
...
 
 #ifdef AVOID_UB
-#define MTX_HUD_POOL_SIZE_MAX MTX_HUD_POOL_SIZE
+#define MTX_hud_POOL_SIZE_MAX MTX_hud_POOL_SIZE
 #else
-#define MTX_HUD_POOL_SIZE_MAX MTX_HUD_POOL_SIZE - 50
+#define MTX_hud_POOL_SIZE_MAX MTX_hud_POOL_SIZE - 50
```
MegaMech, line 18: Same for these two

MegaMech (changes requested):

MegaMech (approved):

PR #531, opened 2024-01-05
Rename some Object related function and property

coco875:
depend on https://github.com/n64decomp/mk64/pull/530


courses/koopa_troopa_beach/course_data.inc.c
```diff
@@ -8538,7 +8538,7 @@ Gfx d_course_koopa_troopa_beach_dl_16990[] =
     gsSPEndDisplayList(),
 };
 
-// 0x60169b0
+//! 0x60169b0, @todo convert to Gfx
```
Tmcg2, line 8541: This and `d_course_yoshi_valley_unk5` aren't Gfx lists. They are their own, arcane thing. Its something related to matrix manipulation, seemingly related to 3d animation (segualls flapping their wings, the flags in YV flapping in the wind, the penguins running around, etc).

coco875: so why in code they are cast like Gfx ?

Tmcg2: Because for some objects (but not all them) that member is a Gfx list. But for others its this weird, u32 list. There is a real possibility that there are several different object types in the same way there's lots of actor types. Its just that no one has gone through and identified what all the various objects really need/use.

As such, any Gfx* type casting going on is an artifact of decompilation.

coco875: and use as an gfx I mean by that is they are use like Gfx so how even it's skeleton/animate model are use in this way ?

Tmcg2: That my memory serves, `code_80004740` is responsible for handling all the model animation stuff, so if you want to know exactly how its used, you'll have to go digging there.

You'll want to read through the comment at the top of `code_80004740.h`, its my best effort to explain what's going on there. I'm way out of my expertise on this one so that comment may or may not be useful

coco875: ok so I remove it or replace by just "more research are needed to know the type"

Tmcg2: And I have no idea if this stuff is similar to SM64 or not. I have looked at SM64's model animation stuff and while it doesn't look similar to me, but I'm really not really knowledgeable enough to say that with certainty.

coco875: I will just remove the todo and see later code_80004740

MegaMech: If I had to guess. The animation data are frame data. Perhaps something like "frame, some other stuff like x,y,z"

coco875: I think I will search that tomorrow to try understand

src/code_8006E9C0.c
```diff
@@ -65,24 +65,24 @@ void reset_object_variable(void) {
 void func_8006EB10(void) {
     s32 i;
     
-    for (i = 0; i < 128; i++) {
-        D_8018C1B0[i] = -1;
+    for (i = 0; i < gObjectParticle1_SIZE; i++) {
```
MegaMech, line 68: defines should be all caps.

But if you intended this as more like a temporary name because you couldn't think of a better one, then this is fine.

src/code_80071F00.c
```diff
@@ -3080,12 +3080,12 @@ void func_8007963C(s32 objectIndex, s32 playerIndex) {
     func_80072488(objectIndex);
 }
 
-void func_80079700(s32 objectIndex, s32 playerIndex) {
+void update_obj_laikitu_redflag(s32 objectIndex, s32 playerIndex) {
```
MegaMech, line 3083: red_flag

MegaMech (changes requested):

src/code_8006E9C0.c
coco875: it's just a replace of name who also replace const so since the const are link to the variable gObjectParticle keep this name it's fine but if you want we can replace by OBJECT_PARTICLE_X_SIZE if you think it's better

tools/doxygen_syms.md
```diff
@@ -1464,14 +1464,14 @@ Variable Name | Memory Address
 [func_800790E4](@ref func_800790E4) | 0x800790E4
 [func_80079114](@ref func_80079114) | 0x80079114
 [func_800791F0](@ref func_800791F0) | 0x800791F0
-[init_obj_laikitu_redflag_coutdown](@ref init_obj_laikitu_redflag_coutdown) | 0x800792D8
+[init_obj_laikitu_red_flag_coutdown](@ref init_obj_laikitu_red_flag_coutdown) | 0x800792D8
```
MegaMech, line 1467: flag_countdown

MegaMech (changes requested):

MegaMech (approved):

PR #538, opened 2024-01-11
Fixes the error "iconv: invalid option -- o" by redirecting stdout to the desired file. 

alyxdeburca:
Was having issues building on MacOS (presumably due to some weird Apple specific version of iconv) but changing this line allowed the build to complete successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -564,7 +564,7 @@ COURSE_DATA_TARGETS := $(foreach dir,$(COURSE_DIRS),$(BUILD_DIR)/$(dir)/course_d
 
 $(BUILD_DIR)/%.jp.c: %.c
 	$(call print,Encoding:,$<,$@)
-	iconv -t EUC-JP -f UTF-8 $< > $@
+	iconv -t EUC-JP -f UTF-8 $< -o $@
 
 $(BUILD_DIR)/%.o: %.c
 	$(call print,Compiling:,$<,$@)
--- a/Makefile
+++ b/Makefile
@@ -564,17 +564,8 @@ COURSE_DATA_TARGETS := $(foreach dir,$(COURSE_DIRS),$(BUILD_DIR)/$(dir)/course_d
 #==============================================================================#
 $(BUILD_DIR)/%.jp.c: %.c
 	$(call print,Encoding:,$<,$@)
-<<<<<<< HEAD
-  ifeq ($(UNAME_S), Darwin)
-	  iconv -t EUC-JP -f UTF-8 $< > $@
-  endif
-  ifeq ($(UNAME_S), Linux)
-    iconv -t EUC-JP -f UTF-8 $< -o $@
-  endif
-=======
 	iconv -t EUC-JP -f UTF-8 $< > $@
 
->>>>>>> parent of 2ec090f80 (Revert "Fixes the error "iconv: invalid option -- o" by redirecting stdout to the desired file. Fixes build errors on macos")
 $(BUILD_DIR)/%.o: %.c
 	$(call print,Compiling:,$<,$@)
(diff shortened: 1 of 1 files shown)


Tmcg2:
Looks like BSD iconv doesn't have the `-o` option at all, which would explain why It doesn't work on MacOS. Good find.

alyxdeburca:
I'll keep checking for more weird MacOS only behaviours

.gitignore
```diff
@@ -82,6 +82,7 @@ expected/*
 !/sound/**/*custom*/**/*.aiff
 !/assets/**/*custom*.bin
 !/assets/**/*custom*/**/*.bin
-
+tools/*.md
 # Doxygen
-docs/html/*
+tools/doxygen_syms.md
+tools/doxygen_syms.md
```
MegaMech, line 88: Please revert this change. We need doxygen_syms generated in the PRs for the github pages API site.

MegaMech (changes requested):

MegaMech:
Could someone test on Linux?

alyxdeburca:
Spinning up a vm now :)

.gitignore
alyxdeburca: Yeah VSCode autmatically did that. Reverting now

alyxdeburca:
I fucked up my local branch. Feel free to fix any git related issues with the last few commits

MegaMech:
> I fucked up my local branch. Feel free to fix any git related issues with the last few commits

Generally, creating a new branch and doing changes in there helps prevent issues with git. And then the master branch is never modified and just synced with the base repo

Makefile
```diff
@@ -561,11 +562,14 @@ COURSE_DATA_TARGETS := $(foreach dir,$(COURSE_DIRS),$(BUILD_DIR)/$(dir)/course_d
...
 $(BUILD_DIR)/%.jp.c: %.c
 	$(call print,Encoding:,$<,$@)
-	iconv -t EUC-JP -f UTF-8 $< -o $@
-
+  ifeq ($(UNAME_S), Darwin)
+	  iconv -t EUC-JP -f UTF-8 $< > $@
+  endif
+  ifeq ($(UNAME_S), Darwin)
+    iconv -t EUC-JP -f UTF-8 $< -o $@
+  endif
```
MegaMech, line 572: I think you might not work Darwin on the second one here? the -o should be non-mac and the > should be mac right?

MegaMech (changes requested):

Makefile
alyxdeburca: Oh yeah, did I copy and pasted and forgot to change that part to Linux instead of darwin

.gitignore
alyxdeburca: @MegaMech can you confirm that this reverted? It's not showing right for me

MegaMech:
I forgot that if the builder passes then it does work on Linux. My bad.

MegaMech:
Is it possible the > works on Linux as well as mac? If it works on both...

MegaMech:
Ya, you're original PR did pass.

alyxdeburca:
My git is so fucked up. I have no idea if this is pushing right

MegaMech:
You could try starting over.

Or try editing the file right in github. You should be able to fix it that way pretty easily.

Click > Files Changed. Find the makefile --> click the three dots --> click edit.

For some reason you have some git conflict merge artifacts in your PR now.

MegaMech:
I fixed it for using in the browser using the edit button.

Lets see if the builder passes

MegaMech (approved):

MegaMech:
Hopefully all the other platforms support this command...

PR #559, opened 2024-01-22
fix warning, type and document audio

coco875:


MegaMech (dismissed):

MegaMech:
lmk when ready for merge. Just thought I'd ask in-case you wanted to add more.

coco875:
I think it will be enough for this PR you can merge it after check last change

src/audio/load.c
```diff
@@ -103,7 +106,7 @@ char loadAudioString38[] = "---------------------------------------\n";
 /**
  * Performs an immediate DMA copy
  */
-void audio_dma_copy_immediate(uintptr_t devAddr, void *vAddr, size_t nbytes) {
+void audio_dma_copy_immediate(u8 *devAddr, void *vAddr, size_t nbytes) {
```
MegaMech, line 109: Are you sure this shuold be u8 * and not uintptr_t?

uintptr_t should be used when doing pointer math. If there's no ptr math then it's probably fine.

src/audio/load.c
```diff
@@ -375,17 +378,17 @@ void func_800BB43C(ALSeqFile *f, u8 *base) {
 
 void patch_sound(struct AudioBankSound *sound, u8 *memBase, u8 *offsetBase) {
     struct AudioBankSample *sample;
-    void *patched;
```
MegaMech, line 378: `void *` should be used if the data is unknown. `u8*` is applicable if the data is a buffer of bytes. Or just a bunch of bytes.

For instances the actor casts in actors.c could arguably be a void * instead of Actor. Ex.
```c
void myActorUpdate(void * actor) {
   MyActor *myActor = (MyActor *) actor;
}
```

src/audio/port_eu.c
```diff
@@ -256,25 +256,23 @@ void func_800CBBE8(u32 arg0, s8 arg1) {
...
-  osSendMesg(D_800EA3AC, thing, 0);
-  D_800EA3A4[0] = D_800EA3A0[0];
+void func_800CBC24(void){
+    s32 temp_t6;
+    s32 test;
+    OSMesg thing;
+    temp_t6 = D_800EA3A0[0] - D_800EA3A4[0];
+    test = (u8) temp_t6;
+    test = (test + 0x100) & 0xFF;
+    do
```
MegaMech, line 266: If this matches can you try (all on one line. Just a little cleaner)

`do {} while(0);`

MegaMech (changes requested):

src/audio/load.c
coco875: I make it to have less errors possible and because all data gives as this argument are u8*

MegaMech: Took a look. Yea, this is fine.

log.txt
MegaMech, line 1: Probably want to remove this file before merging?

MegaMech (approved):

MegaMech (approved):

PR #574, opened 2024-02-15
Match the last 2 data files

Tmcg2:
The names `some_data` and `some_data_2` are obviously terrible but I don't know what else to call them. I am open to any and all suggestions for improved names.


src/data/some_data_2.c
```diff
@@ -0,0 +1,377 @@
+#include "data/some_data_2.h"
+#include "courses/all_course_data.h"
+#include "ending/ceremony_data.inc.h"
+
```
MegaMech, line 4: Name this file `path_spawn_data.c`?

MegaMech (changes requested):

MegaMech (dismissed):

MegaMech:
```c
src/data/some_data.c:185:14: fatal error: assets/onomatopoeia/gTLUTOnomatopoeia.inc.c: No such file or directory
     #include "assets/onomatopoeia/gTLUTOnomatopoeia.inc.c"
```

Tmcg2:
Think I fixed it, I had an incorrect path in onomatopoeia asset makefile that was likely causing the tlut .inc.c file to not get generated

MegaMech (approved):

PR #577, opened 2024-02-26
extract models with blender

coco875:


courses/koopa_troopa_beach/course_data.inc.c
```diff
@@ -8,13 +8,14 @@
...
 #include <courses/koopa_troopa_beach/course_displaylists.inc.h>
+#include <courses/koopa_troopa_beach/course_textures.linkonly.h>
 
 Gfx d_course_koopa_troopa_beach_dl_0[] =
 {
     gsDPTileSync(),
     gsDPSetTile(G_IM_FMT_RGBA, G_IM_SIZ_16b, 8, 0x0000, G_TX_RENDERTILE, 0, G_TX_NOMIRROR | G_TX_WRAP, 5, G_TX_NOLOD, G_TX_NOMIRROR | G_TX_WRAP, 5, G_TX_NOLOD),
     gsDPSetTileSize(G_TX_RENDERTILE, 0, 0, 0x007C, 0x007C),
-    gsDPSetTextureImage(G_IM_FMT_RGBA, G_IM_SIZ_16b, 1, 0x05003800),
+    gsDPSetTextureImage(G_IM_FMT_RGBA, G_IM_SIZ_16b, 1, gKTBTexture660D8C),
```
MegaMech, line 18: We should find a better naming convention for course textures.

Perhaps something like tex_ktb660D8C.

or maybe it makes sense to do some wacky abbreviations here like:
ktb_tex_660D8C.

I can merge as is for now if you don't want to go through this effort now.

However, `g` for global imo should only be used for actual game code and not for symbols that contain textures and the like.

tools/actualise_type/main.py
MegaMech, line 1: Explain why delete please. Not used right?

MegaMech:
Could we put all the blender related files in a folder like:
`blender/myscript.py` and `blender/3D_models.json`

courses/koopa_troopa_beach/course_data.inc.c
coco875: We can let like this but I will some rename of texture and model before merging it

tools/actualise_type/main.py
coco875: Not used

coco875:
By putting the blender extension in the repo, script can interact with it without problem

open_blender_fast64.bat
```diff
@@ -0,0 +1 @@
+"c:\Program Files\Blender Foundation\Blender 4.0\blender.exe" --python tools/blender_extension/fast64_run.py
```
MegaMech, line 1: This is something that's going to break when Blender updates.

Can we have remove these three shortcut files and instead have it in the makefile like for the emulator?

`make blender` or `make open_blender` or something?

MegaMech (changes requested):

open_blender_fast64.bat
MegaMech: And let me know your thoughts on this idea just in-case this idea isn't good.

coco875: to be more precise this file are not planned to stay in the final version it's just to test it can be open_blender_fast64 it's fine but fast64 can work between 3.2-4.0 so just precise the version will break with other version avaible so I will make under another way

MegaMech: It's fine if you hit the latest or most versions. Doesn't need to be perfect.

MegaMech:
I just want to have the chance to test this PR. Then I'll merge.
Or if Taggerung wants to test it.

Just like to have a second person try it out.

coco875:
need some tweak to some json and then it's good

coco875:
need https://github.com/Fast-64/fast64/pull/440 to be merge on fast64 but else it work

coco875:
ready to merge

PR #586, opened 2024-03-21
Rename cpu behaviour related symbols

coco875:


src/code_80005FD0.c
```diff
@@ -91,17 +91,17 @@ u16 D_80163240[12];
 u16 D_80163258[12];
 u16 D_80163270[12];
 s32 D_80163288[10];
-UnkCommonTextureStruct0 *D_801632B0;
-u16 D_801632B8[12];
-u16 D_801632D0[12];
-u16 D_801632E8[12];
+CPUBehaviour *gActualCPUBehaviour;
```
MegaMech, line 94: `gCurrentCPUBehaviour`?

src/code_80005FD0.c
```diff
@@ -3662,7 +3662,7 @@ void func_8000F628(void) {
             }
         }
     }
-    func_80011A5C();
+    copy_coursesc_cpu_behaviour();
```
MegaMech, line 3665: `copy_courses_`

src/common_textures.h
```diff
@@ -5,10 +5,25 @@
...
+    BEHAVIOUR_3,
+    BEHAVIOUR_4,
+    BEHAVIOUR_5,
+    BEHAVIOUR_NORMAL_SPEED,
+    BEHAVIOUR_FAST_SPEED,
+    BEHAVIOUR_SLOW_SPEED,
+    BEHAVIOUR_9,
+    BEHAVIOUR_10,
+    BEHAVIOUR_MAX_SPEED
+};
```
MegaMech, line 26: I think it's best practice to name Enums. And then you can do something like
```c
enum CPUBehaviour {
    NONE,
    FAST,
    NORMAL,
};
```
Without having to repeat `BEHAVIOUR_` for every enum.

src/common_textures.h
```diff
@@ -232,23 +247,23 @@ extern Gfx D_0D008E48[];
...
-extern UnkCommonTextureStruct0 D_0D009188[];
-extern UnkCommonTextureStruct0 D_0D0091E8[];
-extern UnkCommonTextureStruct0 D_0D009210[];
-extern UnkCommonTextureStruct0 D_0D009238[];
-extern UnkCommonTextureStruct0 D_0D009260[];
-extern UnkCommonTextureStruct0 D_0D009280[];
-extern UnkCommonTextureStruct0 D_0D0092C8[];
-extern UnkCommonTextureStruct0 D_0D009310[];
-extern UnkCommonTextureStruct0 D_0D0093C0[];
+extern CPUBehaviour common_cpu_behaviour_battle_course[];
```
MegaMech, line 250: `battle_courses`?

MegaMech (changes requested):
Good work on this PR. Thanks for the help!

src/code_80005FD0.c
coco875: image
image
it's more a local variable and it's not per CPU it's just a place where you save the CPUBehaviour but it's not really global instead of gCurrentCPUBehaviourId

src/common_textures.h
coco875: the problem is if I don't put the prefix BEHAVIOUR_ we can't differenciate a FAST to another FAST because they have the same name but not necessarily the same value

src/data/path_spawn_data.c
```diff
@@ -2,13 +2,29 @@
...
 #include "ending/ceremony_data.inc.h"
 
-UnkCommonTextureStruct0 *D_800DC720[] = {
-	D_0D008F28, D_0D008F80, D_0D008FB8, D_0D009058,
-	D_0D0090B8, D_0D0090F8, D_0D009158, D_0D009188,
-	D_0D0091E8, D_0D009210, D_0D009238, D_0D009260,
-	D_0D009280, D_0D0092C8, D_0D009310, D_0D008F18,
-	D_0D008F18, D_0D008F18, D_0D0093C0, D_0D008F18,
-	D_0D008F18, NULL,
+CPUBehaviour *gRomCoursesCPUBehaviour[NUM_COURSES+1] = {
```
MegaMech, line 5: gCPUBehaviourList
gCPUBehaviourLUT (look up table)

`rom` isn't really accurate here.

src/code_80005FD0.c
MegaMech: `sCurrentCPUBehaviour`?

src/common_textures.h
MegaMech: Resolving because what you did is correct.

src/data/path_spawn_data.c
```diff
@@ -2,13 +2,29 @@
...
 #include "ending/ceremony_data.inc.h"
 
-UnkCommonTextureStruct0 *D_800DC720[] = {
-	D_0D008F28, D_0D008F80, D_0D008FB8, D_0D009058,
-	D_0D0090B8, D_0D0090F8, D_0D009158, D_0D009188,
-	D_0D0091E8, D_0D009210, D_0D009238, D_0D009260,
-	D_0D009280, D_0D0092C8, D_0D009310, D_0D008F18,
-	D_0D008F18, D_0D008F18, D_0D0093C0, D_0D008F18,
-	D_0D008F18, NULL,
+KartAIBehaviour *gKartAIBehaviourLUT[NUM_COURSES+1] = {
```
MegaMech, line 5: This is fine, because my PR will change this array completely.

For future reference, don't hard-coded the array size in `[]` because that means the array will not shift.

Albeit, perhaps a good way to get a warning that you forgot to update X array. So now that I think about it, it *could* be useful. But sometimes it's better to just let the array shift if things are added.

MegaMech:
lgtm!

MegaMech (approved):

src/data/path_spawn_data.c
coco875: hmm ok I think a more thought about how unfix fix array are needed but if torch will generate this code it's fine

MegaMech: In C you don't need to put anything in `[]`. C calculates the array size automatically and passes it onto other files so long as you have declared it in the header.

The only time we need to hard-coded the array size is when we can't link 2 files together properly (course textures and course displaylists or whatever file it is for example.)

MegaMech (approved):

src/data/path_spawn_data.c
coco875: ok ok

PR #588, opened 2024-03-28
some rename for torch

coco875:


include/defines.h
```diff
@@ -281,12 +285,6 @@
 #define DEMO_SIX   5 // Mario Luigi Peach Toad, Sherbert Land,   Versus
 #define NUM_DEMOS  6
 
-#define CUP_COURSE_ONE      0
-#define CUP_COURSE_TWO      1
-#define CUP_COURSE_THREE    2
-#define CUP_COURSE_FOUR     3
-#define NUM_COURSES_PER_CUP 4
-
```
MegaMech, line 289: Should num courses per cup be `NUM_CUPS - 1`?

src/code_80091750.c
```diff
@@ -486,11 +486,11 @@ char *gDebugCourseNames[] = {
...
-    3, 2, 3, 2,
-    1, 1, 2, 2,
-    0, 1, 0, 3,
-    1, 3, 0, 1,
-    3, 2, 0, 0,
+    SPECIAL_CUP, STAR_CUP, SPECIAL_CUP, STAR_CUP,
+    FLOWER_CUP, FLOWER_CUP, STAR_CUP, STAR_CUP,
+    MUSHROOM_CUP, FLOWER_CUP, MUSHROOM_CUP, SPECIAL_CUP,
+    FLOWER_CUP, SPECIAL_CUP, MUSHROOM_CUP, FLOWER_CUP,
+    SPECIAL_CUP, STAR_CUP, MUSHROOM_CUP, MUSHROOM_CUP,
```
MegaMech, line 493: Yesterday I noticed something doesn't seem right about this array.

Choco mountain is course id 1 which we'd expect as flower cup here. But if 2 = star cup. That seems weird. This might not be cupIndexByCourseId but perhaps cupIndexByCupId or something weird like that.

src/data/path_spawn_data.c
```diff
@@ -349,27 +349,27 @@ BombKartSpawn D_800DCC08[][NUM_BOMB_KARTS_MAX] = {
 };
 
 struct _struct_D_800DD9D0_0x10 D_800DD9D0[] = {
-    { 0x0258, 0x0001, 0x0001, 0x0001, 0x0001, { 0x0000, 0x0000, 0x0000}},
```
MegaMech, line 352: This struct contains the size of the course path.

So for instance, Yoshi's valley has (iirc) 4 paths. So it should have for sizes for that course. If you measure the size of the course path in course_data.inc.c you'll find them to be the same. I don't remember what the other values do.

However, they are all zero. So presumably they do nothing.

src/spawn_players.c
```diff
@@ -465,7 +465,7 @@ void func_80039DA4(void) {
         4, 5, 6, 7,
     };
 
-    if (((gCupCourseSelection == CUP_COURSE_ONE) && (D_8016556E == 0)) ||
+    if (((gCupCourseSelection == MUSHROOM_CUP) && (D_8016556E == 0)) ||
```
MegaMech, line 468: Could you verify this that this is supposed to be the cup name rather than the course in the cup?

MegaMech (changes requested):

src/menus.c
```diff
@@ -1718,26 +1718,26 @@ void course_select_menu_act(struct Controller *arg0, u16 arg1) {
                 } else {
                     D_8018EDEC = 3;
                     play_sound2(SOUND_MENU_SELECT);
-                    gCurrentCourseId = gCupCourseOrder[gCupSelection][CUP_COURSE_ONE];
+                    gCurrentCourseId = gCupCourseOrder[gCupSelection][MUSHROOM_CUP];
```
MegaMech, line 1721: Good catch.

include/defines.h
```diff
@@ -156,12 +156,27 @@
...
+
+/**
+ * @brief Options for gCourseIndexInCup
+ */
+enum {
+    FIRST_COURSE_IN_CUP,
+    SECOND_COURSE_IN_CUP,
+    THIRD_COURSE_IN_CUP,
+    FOURTH_COURSE_IN_CUP
+};
```
MegaMech, line 178: Could we keep `CUP_COURSE_ONE` etc. ?

MegaMech (changes requested):

src/menus.c
coco875: I forget to change this one

include/defines.h
coco875: for me FIRST_COURSE_IN_CUP are more clear then CUP_COURSE_ONE because I understand CUP_COURSE_ONE as the first CUP

MegaMech: It doesn't really fit the naming convention though

src/menus.c
MegaMech: Ah, yea you're right.

include/defines.h
coco875: if you want we can remove the IN key word but it's just to be more precise of what it is

coco875: just replace by COURSE_ONE, COURSE_TWO, COURSE_THREE, COURSE_FOUR so just remove the prefix CUP_

MegaMech: That's fine

MegaMech (approved):

PR #594, opened 2024-04-04
Add some comments about the tkmk00 stuff

Tmcg2:


asm/tkmk00_decode.s
```diff
@@ -445,6 +461,7 @@ glabel func_80040AC8
 /* 0417B8 80040BB8 03E00008 */  jr    $ra
 /* 0417BC 80040BBC 00000000 */   nop   
 
+# func_80040BC0 -> SetUpHuffmanTree
```
coco875, line 464: why not rename it ?

asm/tkmk00_decode.s
```diff
@@ -10,6 +10,20 @@
...
+https://github.com/mib-f8sm9c/MiscellaneousHacks/blob/master/BetaTools/MarioKartTestingTool/TKMK00Encoder.cs
+https://skelux.net/showthread.php?tid=47&page=8
+
+This link appears to have code near identical to libtkmk000
+https://github.com/mib-f8sm9c/MiscellaneousHacks/blob/master/BetaTools/MarioKartTestingTool/TKMKTest.cs
+Did libtkmk00 pull from that link, or vice-versa?
+
+Maybe tkmk00 wasn't compiled with IDO. Its explicit usage of registers rather than variables is something that IDO can't do (I think?) but GCC can.
+Or mabybe its C++ code and that interacts with registers differently?
+Or maybe its all hand written?
```
MegaMech, line 24: It's my understanding that this assembly was hand-written.

MegaMech (changes requested):

MegaMech:
Taking another look at this PR.

Everyone whose looked at this func said it was hand-written. My only issue is the comment about it being possibly compiled. It's my understanding that there's no known compiler that outputs assembly like this.

asm/tkmk00_decode.s
maraakate: It is if you know where to look.

MegaMech: The code was likely assembled using IDO, however the code is almost certainly hand-written given the bal, add, and sub, instructions which are not used by any known compiler.

PR #596, opened 2024-04-04
Make trig tables shiftable

MegaMech:
This PR is from the GCC PR.

It should be its own PR so that the improvement can be used now. Also, organizes things a bit.

How this works:

random_seed related variables are moved to the racing segment just after memory.h(.bss). This allows the beginning of the buffers segment to be the trig tables. This allows the trig tables to move correctly with the fixes in segments.h

This allows the .data trig tables segment (the actual data) to shift by referencing the `_buffersSegmentNoloadStart` linker symbol.


coco875:
it's not a PR for uk64 ?

MegaMech:
We need it for uk64 as well!

PR #598, opened 2024-04-09
fix a lot of warning in code

coco875:


coco875:
with using `make 2> log.log` we switch from 7529 to 358. error left are :
- comparison is always true and comparison is always false
- Warning: -Xfullwarn not supported.
- member cannot be of function or incomplete type. from ido (doing it avoid doing multiple struct with arrays of a different size)
- ignoring ‘#pragma etc
- left-hand operand of comma expression has no effect
- function returns address of local variable

courses/courseTable.h
```diff
@@ -16,7 +16,7 @@ struct CourseTable {
     CourseVtx *vertexStart;    // 0x18 segmented address of vertex data
     u32 vertexCount;    // 0x1C number of vertices in vertex data
     u8 *packedStart;    // 0x20 packed display list start address
-    uintptr_t finalDisplaylistOffset;        // 0x24
+    u8* finalDisplaylistOffset;        // 0x24
```
MegaMech, line 19: This is *technically* a uintptr_t in that it contains an addr. Is it possible to do uintptr_t * or uintptr_t?

src/audio/load.c
```diff
@@ -384,9 +384,9 @@ void patch_sound(struct AudioBankSound *sound, u8 *memBase, u8 *offsetBase) {
 #define PATCH(x, base) (patched = (void *)((uintptr_t) (x) + (uintptr_t) base))
 
     if (sound->sample != NULL) {
-        sample = sound->sample = (struct AdpcmLoop *) PATCH(sound->sample, memBase);
+        sample = sound->sample = (struct AudioBankSample *) PATCH(sound->sample, memBase);
         if (sample->loaded == 0) {
-            sample->sampleAddr = (struct u8 *) PATCH(sample->sampleAddr, offsetBase);
```
MegaMech, line 389: This is a funny one. struct lol

src/code_8006E9C0.c
```diff
@@ -240,7 +241,7 @@ void func_8006F008(void) {
         D_80165728 = -330;
         break;
     case COURSE_LUIGI_RACEWAY:
-        D_8018D220 = dma_textures(gTextureExhaust2, 0x4F4U, 0xC00);
+        D_8018D220 = (void *)dma_textures(gTextureExhaust2, 0x4F4U, 0xC00);
```
MegaMech, line 244: I'm wondering if `D_8018D220` should be a u8* instead of `void*` type. If it's always a texture that it's applied to.

src/code_80091750.c
```diff
@@ -1515,9 +1516,9 @@ void func_80091B78(void) {
     gNextFreeMemoryAddress = gFreeMemoryResetAnchor;
     // Hypothetically, this should be a ptr... But only hypothetically.
     D_8018D9B0 = get_next_available_memory_addr(0x000900B0);
-    D_8018D9B4 = (uintptr_t *) get_next_available_memory_addr(0x0000CE00);
-    D_8018D9B8 = (uintptr_t *) get_next_available_memory_addr(0x00012C00);
-    D_8018D9C0 = (struct_8018EE10_entry *) get_next_available_memory_addr(0x00001000);
+    D_8018D9B4 = (u8 *) get_next_available_memory_addr(0x0000CE00);
+    D_8018D9B8 = (u8 *) get_next_available_memory_addr(0x00012C00);
```
MegaMech, line 1520: This is the only other concern I had. I think your change is probably fine. Just need a sec to look into this to confirm.

MegaMech (changes requested):

src/code_8006E9C0.c
MegaMech: Took a closer look. Does (u8 *) work here? I think it's more correct in this scenario.

src/code_80091750.c
MegaMech: Took a look, I like the u8 * change

courses/courseTable.h
MegaMech: Perhaps it should be `Gfx *` since it's technically a ptr to some gfx. I just think that's sort of confusing in a sense cause it's not really a ptr to a displaylist. But tomato tomato?

MegaMech: uintptr_t * or Gfx * I'm good with if either of those works.

MegaMech:
Thanks for the work on this PR! It's really good!

courses/courseTable.h
KiritoDv: I think Gfx* is better

warning.log
MegaMech, line 1: Please delete this file.

MegaMech (changes requested):

PR #603, opened 2024-04-10
add GitHub action and fix build

coco875:


coco875:
I also put all command run before build start in a build.sh to avoid run make assets and after make and it also help clean the Makefile

PR #621, opened 2024-04-21
Fix typos

birdoplank:


Makefile
```diff
@@ -1,5 +1,6 @@
 # Makefile to rebuild MK64 split image
 
+
```
coco875, line 3: why add a new-line fix typos ?

Janmm14: The first PR's of this account were introducing curl script download, potential malware into repos and now for whatever reason is spamming a bunch of repositories with single new empty line change PRs

coco875: ok ok

PR #643, opened 2024-06-30
avoid ub on course segment

coco875:


MegaMech:
The avoid_ub compiles right?

MegaMech (approved):

coco875:
Yes but weirdly doesn't match will retry again later

PR #644, opened 2024-06-30
change bool

coco875:
be more accurate to stbbool https://koor.fr/C/cstdbool/cstdbool.wp


MegaMech (approved):

MegaMech:
Thanks for making this change!

PR #647, opened 2024-07-06
some rename in ceremony

coco875:


src/ending/ceremony_and_credits.c
```diff
@@ -122,80 +122,80 @@ s32 f32_lerp(f32 *dest, f32 src, f32 lerp) {
     return 1;
 }
 
-UNUSED s32 func_80282200(Vec3s arg0, s16 arg1, s16 arg2) {
-    s16 temp_v0 = *arg0;
+UNUSED bool ease_out_transition(s16* var, s16 goal, s16 speed_factor) {
```
MegaMech, line 125: `dest` instead of `goal`?

src/ending/ceremony_and_credits.c
```diff
@@ -122,80 +122,80 @@ s32 f32_lerp(f32 *dest, f32 src, f32 lerp) {
...
+        return FALSE;
     }
-    return 1;
+    return TRUE;
 }
 
 // Calculates fade in/out
-s32 set_transition_colour_fade_alpha_ending(f32 *arg0, f32 arg1, f32 arg2) {
-    f32 temp_f0 = arg1 - *arg0;
+bool adjust_f32_value_transition(f32 *var, f32 goal, f32 speed_factor) {
```
MegaMech, line 144: dest

src/ending/ceremony_and_credits.c
```diff
@@ -205,30 +205,30 @@ void reset_spline(void) {
     D_80287B20 = 0;
 }
 
-void func_80282434(UNUSED struct CinematicCamera *arg0) {
+void wrap_reset_spline(UNUSED CinematicCamera *arg0) {
```
MegaMech, line 208: reset_spline_wrap? wrap as in 'this function is a wrapper for the other function'?

when wrap is in the beginning it means means different

src/ending/ceremony_and_credits.c
```diff
@@ -205,30 +205,30 @@ void reset_spline(void) {
...
     reset_spline();
 }
 
-void func_80282454(Vec3f arg0, Vec3f arg1, f32 *distance, s16 *arg3, s16 *arg4) {
+void calculate_distance_angle_y_and_angle_Y_to_XZ(Vec3f vec1, Vec3f vec2, f32 *distance, s16 *angleYToXZ, s16 *angleY) {
```
MegaMech, line 212: consistent caplitalization. Y_and_angle_Y_to_XZ or y_and_angle_xz.

src/ending/ceremony_and_credits.c
```diff
@@ -205,30 +205,30 @@ void reset_spline(void) {
...
-    *arg4 = atan2s(xdist, zdist);
+    *angleYToXZ = atan2s(ydist, sqrtf((xdist * xdist) + (zdist * zdist)));
+    *angleY = atan2s(xdist, zdist);
 }
 
-void func_80282504(Vec3f arg0, Vec3f arg1, f32 arg2, s16 arg3, s16 arg4) {
-    arg1[0] = arg0[0] + (arg2 * coss(arg3) * sins(arg4));
-    arg1[1] = arg0[1] + (arg2 * sins(arg3));
-    arg1[2] = arg0[2] + (arg2 * coss(arg3) * coss(arg4));
+void apply_distance_angle_y_and_angle_Y_to_XZ(Vec3f vec1, Vec3f vec2, f32 coef, s16 angleYToXZ, s16 angleY) {
```
MegaMech, line 225: capitalization

src/ending/ceremony_and_credits.c
```diff
@@ -205,30 +205,30 @@ void reset_spline(void) {
...
+    vec2[0] = vec1[0] + (coef * coss(angleYToXZ) * sins(angleY));
+    vec2[1] = vec1[1] + (coef * sins(angleYToXZ));
+    vec2[2] = vec1[2] + (coef * coss(angleYToXZ) * coss(angleY));
 }
 
-UNUSED void func_802825C8(Vec3f arg0, Vec3f arg1, Vec3f arg2, Vec3s arg3) {
+UNUSED void avorting_cinematic_function(Vec3f arg0, Vec3f arg1, Vec3f arg2, Vec3s arg3) {
```
MegaMech, line 231: aborting?

src/ending/ceremony_and_credits.c
```diff
@@ -566,191 +564,192 @@ void init_cinematic_camera(void) {
...
-        func_80282454(new_var->lookAt, new_var->pos, &sp64, &sp6E, &sp6C);
-        if (sp6E >= 0x3800) {
-            sp6E = 0x3800;
+    f32 distance;
+    Vec3f pos;
+    Vec3f lookAt;
+    Vec3f up;
+    CinematicCamera *cinematicCamera = &D_802876E0;
+
+    cinematic_empty_0_function();
```
MegaMech, line 591: `cinematic_useless_function`?

MegaMech (changes requested):
Alright this PR is pretty good.

My only concern is the removal of this:
```
typedef BAD_RETURN(s32) (*CameraEvent)(Camera *c);
typedef CameraEvent CutsceneShot;
```
So for further context. SM64 had a decent reason for doing this. But mk64 doesn't really.

I just want to quickly confirm why it's there

coco875:
Hmm normally it's fine I think

src/ending/ceremony_and_credits.c
MegaMech: cinematic_stub()

src/ending/ceremony_and_credits.c
MegaMech: exit_cinematic() ?

src/ending/ceremony_and_credits.c
coco875: when it's before what does it mean for you ?

src/ending/ceremony_and_credits.c
coco875: it's just a function who return 0 and are use, so useless no, and I don't think stub are good also

src/ending/ceremony_and_credits.c
MegaMech: 'wrap reset spline' turns wrap into a verb. Which some might think that this means that the 'reset spline' gets wrapped. Rather than the intention which is to say that this is just a wrapper function.

A couple other ideas to name it. 'reset_spline_helper' 'reset_spline_wrapper' 'reset_spline_wrap' 'reset_spline_decorator'

src/ending/ceremony_and_credits.c
MegaMech: Despite that it's used. It doesn't actually do anything with the return value (At least not that I could see).

I think _stub is suitable to show that there's not really anything interesting here.

PR #648, opened 2024-07-10
update libultra asm

coco875:
use https://github.com/n64decomp/libreultra to update asm of libultra


MegaMech (commented):
My only concern is the thread tail parts where I don't think it will shift properly on gcc.

Makefile
```diff
@@ -33,7 +33,7 @@ $(eval $(call validate-option,COMPILER,ido gcc))
 DEBUG ?= 0
 
 # Compile with GCC
-GCC ?= 0
+GCC ?= 1
```
coco875, line 36: oh forget to set again to 0

PR #655, opened 2024-07-30
fix typo buffers

coco875:


coco875:
match

src/buffers/memory_pool.c
```diff
@@ -8,4 +8,4 @@
  * Memory pool variable prevents code segments flowing into the memory pool
  * for easier portability.
 */
-u8 memoryPool[MEMORY_POOL_SIZE];
+u8 gMemoryPool[MEMORY_POOL_SIZE];
```
MegaMech, line 11: Fair.

But this variable should never actually be used. If anything it should be a static const if that's even possible.

Putting g makes it seem like it could be used. Sould be sMemoryPool at the least.

Or perhaps even sMemoryPoolBss or something (if this is bss)

MegaMech (changes requested):

src/buffers/memory_pool.c
coco875: I rename to sMemoryPool and add a warning
image

MegaMech (approved):

PR #673, opened 2024-09-01
Rename Object Timing Related Code

coco875:


src/update_objects.c
```diff
@@ -7856,7 +7859,7 @@ void func_80086700(s32 objectIndex) {
     set_obj_origin_offset(objectIndex, 0.0f, 0.0f, 0.0f);
     set_obj_direction_angle(objectIndex, 0U, 0U, 0U);
     gObjectList[objectIndex].unk_084[1] = 0x0200;
-    func_80072488(objectIndex);
+    reset_timer_go_next_state(objectIndex);
```
MegaMech, line 7862: `reset_timer_run_next_state`?

MegaMech (changes requested):

src/update_objects.c
MegaMech: or set_next_state

MegaMech: initiate_next_state

coco875: Run_next_state are not correct because it don't call a function

coco875: Initialise_next_state would be better

coco875:
wrong merge before but else it's should be fine

src/update_objects.c
```diff
@@ -489,18 +492,18 @@ s32 func_80072B48(s32 objectIndex, s32 arg1) {
     s32 phi_v1;
 
     phi_v1 = 0;
-    if (gObjectList[objectIndex].unk_0CB == 0) {
-        set_object_unk_0CB(objectIndex, 1);
+    if (gObjectList[objectIndex].isTimerRunning == 0) {
+        set_object_timer_running(objectIndex, 1);
```
MegaMech, line 496: trigger_object_timer ?
or start_object_timer

MegaMech (changes requested):

src/update_objects.c
MegaMech: the name itself is fine. Just thought this might be better. You choose.

coco875: for me trigger_object_timer or start_object_timer will don't get an bool to set and if we put at the 0 it will not start

src/update_objects.c
```diff
@@ -267,8 +267,8 @@ bool func_80072354(s32 objectIndex, s32 arg1) {
     return b;
 }
 
-void set_object_unk_0CB(s32 objectIndex, s32 arg1) {
-    gObjectList[objectIndex].unk_0CB = arg1;
+void set_object_timer_running(s32 objectIndex, s32 arg1) {
+    gObjectList[objectIndex].isTimerRunning = arg1;
```
MegaMech, line 271: isTimerActive ?

src/update_objects.c
MegaMech: `toggle_timer_state` or `toggle_object_timer`?

coco875: for me toggle it's automatically switch from true to false and vice-versa and don't have an extra arg to set it

coco875: hmm set_object_timer_state ?

MegaMech: Sure

src/update_objects.c
MegaMech: initialize

src/update_objects.c
coco875: done

src/update_objects.c
MegaMech: set_next_animation_state ?

MegaMech: advance_object_state?

coco875: The timer it's not for animation it's for object to know when he go to the next state

MegaMech (approved):

PR #677, opened 2024-09-05
Fix some libultra struct types

AloXado320:
Also updated some libultra files alongside adding extra GCC toolchain support


Makefile
```diff
@@ -279,14 +279,20 @@ GLOBAL_ASM_RACING_O_FILES = $(foreach file,$(GLOBAL_ASM_RACING_C_FILES),$(BUILD_
 
 # detect prefix for MIPS toolchain
 ifneq ($(CROSS),)
```
coco875, line 281: why you rearrange the priority ? mips64-elf- it's the version of libdragon no ? on your os it work only with a specific version ?

src/os/__osDequeueThread.c
```diff
@@ -1,23 +1,24 @@
 #include "libultra_internal.h"
 
-NO_REORDER OSThread* __osThreadTail = NULL;
-NO_REORDER u32 __osTest = -1;
-NO_REORDER OSThread* __osRunQueue = (OSThread*) &__osThreadTail;
-NO_REORDER OSThread* __osActiveQueue = (OSThread*) &__osThreadTail;
-OSThread* __osRunningThread = NULL;
-OSThread* __osFaultedThread = NULL;
+struct __osThreadTail __osThreadTail = { NULL, -1 };
```
coco875, line 3: did you update from https://github.com/n64decomp/libreultra/blob/1aca5c13ca041cef86f8dc194b727361dad9c09b/src/os/thread.c ? and if yes why not rename it ?

src/os/osEepromRead.c
```diff
@@ -35,7 +35,7 @@ s32 osEepromRead(OSMesgQueue* mq, u8 address, u8* buffer) {
     ret = __osSiRawStartDma(OS_WRITE, &__osEepPifRam);
     osRecvMesg(mq, NULL, OS_MESG_BLOCK);
 
-    for (i = 0; i < ARRAY_COUNT(__osEepPifRam.ramarray) + 1; i++) {
+    for (i = 0; i < ARRLEN(__osEepPifRam.ramarray) + 1; i++) {
```
coco875, line 38: why use ARRLEN instead of ARRAY_COUNT ?

coco875 (commented):
just few question

Makefile
AloXado320: It uses the same arragement as in this fork of SM64
https://github.com/CrashOveride95/ultrasm64/blob/master/Makefile#L329

src/os/__osDequeueThread.c
AloXado320: I used ultralib, I would leave the renaming for another PR since I just focused on fixing types

src/os/osEepromRead.c
AloXado320: ultralib makes more use of it

Makefile
coco875: yes but why ?

src/os/__osDequeueThread.c
coco875: ok ok

src/os/osEepromRead.c
coco875: ok then there is no problem

Makefile
AloXado320: My WSL2 environment has a different N64 toolchain, I can build IDO just fine but for GCC i need those extra cross checks

coco875: ok ok it's fine for me also

src/os/__osDequeueThread.c
MegaMech: This change regarding threadtail stuff is actually a reversion.

I don't really know which verison is better or not.

AloXado320: It's on a struct so there's no reordering, and it still matches

MegaMech: Does this work on GCC?

MegaMech (approved):

src/os/__osDequeueThread.c
AloXado320: It does boot on GCC, but it still has some problems unrelated to this

PR #700, opened 2025-01-30
document greyscale math

coco875:


MegaMech (approved):

MegaMech (approved):

MegaMech:
I have 2 concerns with this PR:

1) Is it possible a compiler will try to reinterpret pow or ln function calls to hardware calls? Like what can happen with sqrt iirc.
2) Given that this is in menu_items.c might these names conflict elsewhere?
  * What I'm saying is that. This isn't math_util.c
  * So naming them pow, exp, ln, could make things confusing.

I'm not saying I don't like the names. It's just a thought, in-case it's worth considering.

coco875:
Ok will do that later

asm/non_matchings/menu_items/exponential.s
```diff
@@ -5,7 +5,7 @@ glabel D_800F0CD0
 
 .section .text
 
-glabel func_8009195C
+glabel exponential
```
MegaMech, line 8: Should this just be called exponent?

MegaMech (changes requested):

MegaMech (approved):

PR #711, opened 2025-05-12
It's called round, it's called a rainbow round, it's a round that you go

AloXado320:
Better late than never I guess...

Anyways, rounding floats are wonderful if used right, which in this case I didn't and even though we hit 100%, I still couldn't let this error slide, oops.


MegaMech:
PR title not funny enough

PR #718, opened 2025-05-15
Update README.md

arefdsg:


MegaMech:
make -C tools (optional), try if having issues

I personally don't like running the recursive command in the clone command like that. I don't mind having both so people can choose but it might over complicate the readme.

MegaMech (dismissed):

PR #722, opened 2025-06-14
Document projection matrix function

JedGrabman:
Document function for computing a projection matrix. Used for camera and graphics.

A good resource for math details and general context for a projection matrix is [here](https://www.songho.ca/opengl/gl_projectionmatrix.html).

Attempted to preserve exact code functionality, but don't know a good way to confirm that. Goal is to improve readability.
One variable is overwritten and only used elsewhere. It is unknown what it is used for.


JedGrabman:
Fixed basic bug introduced in 1st commit and updated comments / variable names in response to comments on the discord.

PR #730, opened 2025-06-26
Document time trial input replay system

JedGrabman:
Documents how the input replay system works in time trials. This is used for
* Course ghosts (on the raceways)
* Player ghosts
* Replays

`staff_ghosts.c` is the main file touched.

Inputs are saved in a `s32[]` where each entry is a combination of the inputs and how long those inputs were held for. In essence it's "These buttons were pressed and the joystick was in this position. This was the case for X frames".

bits 1-8: Stick X
bits 9-16: Stick Y
bits 17-24: frame counter
bits 25-28: Unused? 
bit 29: R-Trigger
bit 30: Z-Trigger
bit 31: B-Button
bit 32: A-Button

Various things can disable the playback (pausing, collisions, Lakitu, etc.).

Need to add a little more documentation, but otherwise I expect to keep PR at this scope.


src/staff_ghosts.c
```diff
@@ -212,261 +212,275 @@ void func_80005310(void) {
...
+    if (inputs & 1 << 28) {
+        buttons |= R_TRIG;
     }
-    temp_v1 = gControllerEight->buttonPressed & 0x1F0F;
-    gControllerEight->buttonPressed = (phi_v0 & (phi_v0 ^ D_80162DA8)) | temp_v1;
-    temp_v1 = gControllerEight->buttonDepressed & 0x1F0F;
-    gControllerEight->buttonDepressed = (D_80162DA8 & (phi_v0 ^ D_80162DA8)) | temp_v1;
-    D_80162DA8 = phi_v0;
-    gControllerEight->button = phi_v0;
+    buttons_temp = gControllerEight->buttonPressed & 0x1F0F;
```
MegaMech, line 280: image

This can be broken by clicking on the bits at the bottom until there's just one at a time. Then check os_cont.h
So the first bit is 1 which is 
```c
CONT_CARD_ON | CONT_CARD_PULL | CONT_ADDR_CRC_ER |  U_CBUTTONS | R_JPAD | L_JPAD | D_JPAD | U_JPAD | START_BUTTON
```
This will clear any bit that isn't in the above whitelist. If the bit is already unset, it remains unset.

So if the user is pressing the A_BUTTON, then the A_BUTTON will be cleared (turned off). If the user is pressing D_JPAD then it stays pressed. If the user is not pressing D_JPAD its state stays the same (not pressed).

So 0x1F0F can be replaced with the above whitelist.

src/staff_ghosts.c
```diff
@@ -212,261 +212,275 @@ void func_80005310(void) {
...
     } else {
-        phi_v1 = (s16) (temp_v0 | (~0xFF));
+        stickVal = (s16) (stickBytes | (~0xFF));
     }
-    gControllerSeven->rawStickY = phi_v1;
+    gControllerSeven->rawStickY = stickVal;
 
-    if (temp_a0 & 0x80000000) {
-        phi_a2 = A_BUTTON;
+    if (inputs & 1 << 31) {
```
MegaMech, line 326: (note: It appears you have a clear grasp of what's going on, but I'm going to walk myself through the basics here anyway)
This isn't the clearest way to display this.

If I remember correctly, what's happening here is we've read the staff ghost controller button action list. (or whatever you want to call it).

So 0x800000000 and 0x400000 and 0x10000000 is the button presses saved in the file. for like a button, b button, etc.

typedef struct {
    u8 button;
    s8 frameDuration;
    s8 stickY;
    s8 stickX;
} StaffGhost;
// 80160ADC Banana's remaining

This struct is used as a u32 in this function. So to access the button struct member we do `entry & button_location` to get just the button value.

It's clearer to make a define like this:

Please place this somewhere in defines.h
```c
/**
 * Staff ghost controller buttons
 * Each StaffGhost entry is converted to a u32 value
 * This allows access to the button struct member
 */
#define STAFF_GHOST_NO_BUTTON 0
#define STAFF_GHOST_A_BUTTON  (1 << 31) // 0x80000000
#define STAFF_GHOST_B_BUTTON  (1 << 30) // 0x40000000
#define STAFF_GHOST_Z_TRIG    (1 << 29) // 0x20000000
#define STAFF_GHOST_R_TRIG    (1 << 28) // 0x10000000
```

And place the defines in this area. Feel free to rename any of the names I've chosen if you feel you think of a better name.

src/staff_ghosts.c
```diff
@@ -212,261 +212,275 @@ void func_80005310(void) {
...
-    gControllerSeven->buttonPressed = (phi_a2 & (phi_a2 ^ D_80162D98)) | temp_v1;
-    temp_v1 = gControllerSeven->buttonDepressed & 0x1F0F;
-    gControllerSeven->buttonDepressed = (D_80162D98 & (phi_a2 ^ D_80162D98)) | temp_v1;
-    D_80162D98 = phi_a2;
-    gControllerSeven->button = phi_a2;
-    if (D_80162D9C == 0) {
-        D_80162DA0++;
-        D_80162D9C = (s32) (D_80162DA4[D_80162DA0] & 0xFF0000);
+    // Blanks the A, B, Z, R and L buttons
+    buttonsTemp = gControllerSeven->buttonPressed & 0x1F0F;
```
MegaMech, line 339: Comment here is good! Please put a define for 0x1F0F here as well and below here as well.

src/staff_ghosts.c
```diff
@@ -212,261 +212,275 @@ void func_80005310(void) {
...
+void save_player_inputs(void) {
+    s16 buttons;
+    u32 inputs;
+    u32 stickX;
+    u32 stickY;
+    u32 inputCounter;
+    u32 prevInputsWCounter;
+    u32 prevInputs;
+    /* Input file is too long or picked up by lakitu or Out of bounds
+    Not sure if there is any way to be considered out of bounds without lakitu getting called */
```
MegaMech, line 432: iirc, there is some geography that can set oob when driven on. But like you say here, lakitu gets called.

src/staff_ghosts.c
```diff
@@ -507,34 +521,36 @@ void func_80005B18(void) {
...
             if ((gPlayerOne->type & 0x800) == 0x800) {
                 func_80005AE8(gPlayerTwo);
                 func_80005AE8(gPlayerThree);
-            } else {
-                D_80162DEC += 1;
-                if (D_80162DEC >= 0x65) {
-                    D_80162DEC = 0x00000064;
+            }
+            else {
+                UNUSED_VAR2 += 1;
```
MegaMech, line 541: Not really a fan of this style of naming for unused variables, because there's lots of these and someone might try to name one unused_var2 as well.

We usually add the UNUSED tag in front of the variable definition.
And usually something like sUnusedStaffGhostCounter or something along those lines.

This is just a minor nitpick. You can ignore this comment if you like.

src/staff_ghosts.c
```diff
@@ -507,34 +521,36 @@ void func_80005B18(void) {
...
+                if (UNUSED_VAR2 >= 101) {
```
MegaMech, line 542: `> 100` should match

src/staff_ghosts.c
```diff
@@ -507,34 +521,36 @@ void func_80005B18(void) {
...
+                    UNUSED_VAR2 = 100;
```
MegaMech, line 543: Comment bubble above was glitchy for me. In-case it didn't save. I was asking for var `> 100` instead of `>= 101` as that should match there and looks better.

src/staff_ghosts.c
```diff
@@ -564,9 +580,10 @@ void staff_ghosts_loop(void) {
...
         return;
     }
-    if (!D_80162DF0) {
+    if (!gPauseTriggered) {
         func_80005B18();
         return;
     }
-    D_80162DF8 = 1;
+    // This only gets triggered when the game is paused.
+    gReplayCannotSave = 1;
```
MegaMech, line 588: Does it not also get triggered above when oob, or lakitu?

MegaMech (changes requested):

src/staff_ghosts.c
JedGrabman: Done

JedGrabman: Resolving as duplicate

src/staff_ghosts.c
JedGrabman: Done

src/staff_ghosts.c
JedGrabman: Ah, I meant that that specific line of code was only triggered when there's a pause. The variable can be set in other places. I changed the comment to be a little clearer.

JedGrabman:
Thanks! It's a pretty cool replay system once I wrapped my head around it.

I ended up deciding to call the ghosts and replay all replays. Which means the basic replay is now a "Post time trial replay". I'll probably add some documentation for that and rename the `course_ghosts` files to reflect this before finalizing the PR.

src/staff_ghosts.c
MegaMech: Sounds good!

MegaMech (dismissed):

JedGrabman:
Renamed `staff_ghosts` files to `replays` and added a comment header to `replays.c`. I think it's good to go

MegaMech (dismissed):

MegaMech:
Do you mind going one step farther and taking a look at
`D_80162DC0`

If I'm understand this correctly, it checks if the course has changed, and if so resets some stuff.

MegaMech (approved):

MegaMech:
Perfecto!

PR #750, opened 2025-08-28
Player bitflag attributes

JedGrabman:
Defining many player attributes that are bit flags. These include:

* `kartProps`, various properties related to the current state of the kart  
* `kartGraphics` related to graphics that appear while driving (e.g. the POOMP! effect when landing)  
* `lakituProps` related to states when being retrieved by lakitu
* `oobProps` related to passing the Out of Bounds (OOB) plane

Less thorough documentation of `UNK_002`, which has something to do with how other players appear to you.

See `include/defines.h` for specific flag values and additional documentation.

### DIFF
--- a/src/audio/external.c
+++ b/src/audio/external.c
@@ -2688,7 +2688,7 @@ void func_800C76C0(u8 playerId) {
 }
 
 void func_800C847C(u8 playerId) {
-    if ((gPlayers[playerId].unk_0DE & 1) == 1) {
+    if ((gPlayers[playerId].unk_0DE & UNK_0DE_UNDER_OOB_OR_FLUID_LEVEL) == UNK_0DE_UNDER_OOB_OR_FLUID_LEVEL) {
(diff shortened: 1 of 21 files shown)


src/audio/external.c
```diff
@@ -1824,7 +1824,7 @@ void func_800C5CB8(void) {
 }
 
 void func_800C5D04(u8 playerId) {
-    if ((gPlayers[playerId].unk_044 & 0x20) == 0x20) {
+    if ((gPlayers[playerId].kartProps & PRESS_A) == PRESS_A) {
```
MegaMech, line 1827: MOVE_FORWARDS?

MegaMech (changes requested):

src/audio/external.c
JedGrabman: Renamed to `THROTTLE` to describe Kart state. I don't believe it requires forward movement.

PR #751, opened 2025-09-21
Add Basic Rumble Support

NewGBAXL:
Included is a sample implementation of Rumble Pak support for up to 4 players.

It uses the general structure of SM64's Shindou update with updated code to properly handle multiple controllers separately.

Current code rumbles on menu select and boosting, but other in-game events can easily be added in the future with hardware testing.

### DIFF
--- a/src/save.c
+++ b/src/save.c
@@ -44,15 +44,28 @@ void func_800B45E0(s32 arg0) {
         &gSaveData.allCourseTimeTrialRecords.cupRecords[arg0 / 4].courseRecords[arg0 % 4];
 
     courseTimeTrialRecordsPtr->checksum = checksum_time_trial_records(arg0);
+#if ENABLE_RUMBLE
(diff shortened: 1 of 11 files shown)


src/save.c
```diff
@@ -44,15 +45,28 @@ void func_800B45E0(s32 arg0) {
         &gSaveData.allCourseTimeTrialRecords.cupRecords[arg0 / 4].courseRecords[arg0 % 4];
 
     courseTimeTrialRecordsPtr->checksum = checksum_time_trial_records(arg0);
+#if ENABLE_RUMBLE
+    //block_until_rumble_pak_free();
```
MegaMech, line 49: Why the if with commented out funcs?

NewGBAXL:
I'd gladly finish the implementation, however I do not have a physical Rumble Pak (yet) so I can't really fine tune it atm.
(It has been tested on original hardware by others)

PR #761, opened 2026-04-10
Steering

JedGrabman:
Document the steering process.

Basically, your desired steering position is set by your joystick's position. However, your steering can only change so much in one frame. 

I recommend starting by reading the large comment in `src/player_controller.c` that explains it in more detail (before many calls to `update_steering_large` and `update_steering_small`)

Note that `player->steerPosition` rarely uses the rightmost 16 bits, so there's a fair amount of bit shifting.

### DIFF
--- a/src/player_controller.c
+++ b/src/player_controller.c
@@ -3901,12 +3901,12 @@ void update_steering_small(Player* player, s32* desired_steering_change, s32* cu
 
 //On or near ground
 void func_80033AE0(Player* player, struct Controller* controller, s8 playerIndex) {
-    s32 steer_angle;
+    s32 steer_position;
     s32 clamped_x;
     UNUSED s32 pad[2];
     UNUSED s16 pad2;
     s16 var_s1_2;
-    s32 steer_angle_delta;
+    s32 steer_position_delta;
     s32 steer_resistance_small_turn;
     s32 steer_resistance_large_turn;
     f32 var_f2_2;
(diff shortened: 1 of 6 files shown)


src/player_controller.c
```diff
@@ -3817,21 +3817,25 @@ void func_80033850(Player* arg0, f32 arg1) {
...
-        temp_v1 = player->unk_200;
-        player->unk_200 -= 0x800;
-        if (player->unk_200 >= 0xF0000000) {
-            player->unk_200 = temp_v1;
+    // if (abs(desired_steering_change) >= steering_change_threshold)
+    if ((*desired_steering_change >= steering_change_threshold) || (-steering_change_threshold >= *desired_steering_change)) {
+        current_steering_change_increment = player->steerChangeIncrement;
+        player->steerChangeIncrement -= 1 << 11;
+        //signed vs unsigned weirdness, actually checking if negative
+        if (player->steerChangeIncrement >= 0xF0000000) {
```
MegaMech, line 3828: Do you think this might technically be UB?

I wonder if it's a compiler optimization and that it might actaully be < 0

src/player_controller.c
```diff
@@ -3974,144 +3982,156 @@ void func_80033AE0(Player* player, struct Controller* controller, s8 arg2) {
...
+        }
+        // Driving in water?
+        if ((((f64) (D_801652A0[playerIndex] - player->tyres[BACK_LEFT].baseHeight)) >= 3.5) ||
+            (((f64) (D_801652A0[playerIndex] - player->tyres[BACK_RIGHT].baseHeight)) >= 3.5)) {
+            steer_resistance_large_turn *= 1.05;
+            steer_resistance_small_turn *= 1.05;
+        }
+    }
+
+    // This big group of function calls does 2 things.
```
MegaMech: Could you please put this comment block in a 

/**
 *
 *
 */
style?

MegaMech (dismissed):

src/player_controller.c
JedGrabman: It's strange that it's checking for `0xF`... instead of `0x8`... If it was compiler checking for < 0, I would expect it to only be checking the first bit for the sign. I'll admit I missed that initially. But it's a detail that makes me think it was coded by a human instead of a compiler.

I think that leads to a theoretical bug. A negative number with a large enough magnitude fails the comparison when I run it on an online c compiler, presumably because the two's complement actually makes it smaller when interpreted as an unsigned int. But, that bug can't actually come into play because `steerChangeIncrement` only changes by 2**11 at a time.

I'm not an expert on undefined behavior, but the wikipedia article on it uses underflows in C as an example, so I suppose it is.

src/player_controller.c
JedGrabman: Done

MegaMech (approved):

PR #766, opened 2026-07-05
Document versus rankings and finish info

JedGrabman:
Documenting variables and behavior related to rankings and finishing.

This was related to an investigation into port priority for versus mode, so added some comments about how that works since it is not immediately obvious by looking at the code. Roughly, player movement gets processed in player order, so if two players finish on the same frame in versus, the player with the lower port number will be processed first and get the better placement.

### DIFF
--- a/src/racing/race_logic.c
+++ b/src/racing/race_logic.c
@@ -81,7 +81,7 @@ void func_8028E028(void) {
             break;
     }
     func_800CA118((u8) gPlayerWinningIndex);
-    D_800DC510 = 5;
+    gRaceState = RACE_DONE;
     gDemoTimer = 10;
 }
 
@@ -225,7 +225,7 @@ void func_8028E438(void) {
                     func_80019DF4();
                 } else {
                     func_80092564();
-                    D_800DC510 = 7;
+                    gRaceState = RACE_QUADRANT_RESULTS;
                 }
             }
             break;
@@ -399,7 +399,7 @@ void func_8028E678(void) {
         case 4:
             gIsInQuitToMenuTransition = 1;
             gQuitToMenuTransitionCounter = 5;
-            D_800DC510 = 7;
+            gRaceState = RACE_QUADRANT_RESULTS;
             func_8028E3A0();
             break;
     }
@@ -407,7 +407,7 @@ void func_8028E678(void) {
 
 UNUSED void func_8028EC38(s32 arg0) {
     gGotoMode = arg0;
-    D_800DC510 = 6;
+    gRaceState = RACE_QUITTING;
     func_800CA330(25);
     func_800CA388(25);
     D_800DC5B4 = 1;
@@ -497,8 +497,8 @@ void start_race(void) {
         play_music_for_current_track(gCurrentCourseId);
     }
 
-    if (D_800DC510 == 2) {
-        D_800DC510 = 3;
+    if (gRaceState == RACE_STAGING) {
+        gRaceState = RACE_CURRENT;
     }
 
     for (i = 0; i < NUM_PLAYERS; i++) {
@@ -562,7 +562,7 @@ void func_8028EF28(void) {
                         D_802BA048 = 1;
                     }
                     if ((player->type & PLAYER_INVISIBLE_OR_BOMB) == 0) {
-                        D_800DC510 = 4;
+                        gRaceState = RACE_HAS_HUMAN_FINISHER;
                     }
                     if (gModeSelection == TIME_TRIALS) {
                         func_80005AE8(player);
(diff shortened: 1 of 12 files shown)


src/racing/race_logic.c
```diff
@@ -497,8 +497,8 @@ void start_race(void) {
         play_music_for_current_track(gCurrentCourseId);
     }
 
-    if (D_800DC510 == 2) {
-        D_800DC510 = 3;
+    if (gRaceState == RACE_STAGING) {
+        gRaceState = RACE_CURRENT;
```
MegaMech: RACE_IN_PROGRESS?

src/racing/race_logic.c
```diff
@@ -571,70 +571,80 @@ void func_8028EF28(void) {
...
                         }
                         switch (gPlayerCountSelection1) {
                             case 2:
                                 if (currentPosition == 0) {
-                                    *(gNmiUnknown1 + i) += 1;
+                                    *(gVersusResults2P + playerId) += 1;
                                 }
-                                if (*(gNmiUnknown1 + i) > 99) {
-                                    *(gNmiUnknown1 + i) = 99;
+                                if (*(gVersusResults2P + playerId) > 99) {
```
MegaMech: Much better name,

however, wondering if it's worth doing something like nmi_gVersusResults2P just to let people know this is using the NMI buffer which is a special memory on the N64 which keeps state when you press the console reset button. Or perhaps a comment... Also check that the variable isn't multi-purpose.

src/racing/race_logic.c
```diff
@@ -571,70 +571,80 @@ void func_8028EF28(void) {
...
+                                    *(gVersusResults2P + playerId) = 99;
                                 }
-                                D_800DC510 = 5;
-                                i = gPlayerPositionLUT[1];
-                                gPlayers[i].triggers |= SPINOUT_TRIGGER;
-                                gPlayers[i].type |= PLAYER_CPU;
-                                func_800CA118((u8) i);
+                                gRaceState = RACE_DONE;
```
MegaMech, line 584: RACE_FINISHED ? (I guess this is a pretty symantic/pedantic suggestion)

src/racing/race_logic.c
```diff
@@ -652,32 +662,32 @@ void func_8028F3E8(void) {
...
     s32 i;
 
-    switch (D_800DC510) {
-        case 3:
-        case 4:
-        case 5:
-        case 7:
+    switch (gRaceState) {
+        case RACE_CURRENT:
+        case RACE_HAS_HUMAN_FINISHER:
```
MegaMech: finisher or finished?

src/racing/race_logic.c
```diff
@@ -522,47 +522,47 @@ f32 func_8028EE8C(s32 arg0) {
     return gCourseTimer - ((COURSE_TIMER_ITER_f * temp_f14) / (temp_f14 + temp_f16));
 }
 
-void func_8028EEF0(s32 i) {
+void add_cinematic_mode(s32 i) {
```
MegaMech, line 525: This is probably fine but wanted to also suggest: `set_player_cinematic_mode`?

MegaMech (changes requested):

src/racing/race_logic.c
JedGrabman: Because you can only enable cinematic mode, but not disable it, I would prefer to avoid using `set`.

src/racing/race_logic.c
JedGrabman: Done

src/racing/race_logic.c
JedGrabman: Done

src/racing/race_logic.c
JedGrabman: Denotes at least 1 human has finished. 

Changed it to `RACE_HUMAN_FINISHED`.  I think `HAS` made it sound more like a boolean than intended.

src/racing/race_logic.c
JedGrabman: I think it's good to be distinct from the idea of finishing a race. Especially since the race can end before everyone finishes.

src/racing/race_logic.c
MegaMech: I don't mind the has, up to you on this one. I just thought the `finisher` was a mispelling and that the intention may have been `finished`

RACE_HAS_A_PLAYER_FINISHED is another alternate option...

But this is also state so maybe:
`RACE_A_PLAYER_FINISHED`

I'm kinda up for any of the above on this one.

MegaMech (approved):

PR #771, opened 2026-07-28
Annotate the always-true condition in func_802B4104

quarrel07:
Documents one vanilla condition in `func_802B4104` (blue/red shells) that never does what it looks like it does. Both branches use `||` where an `&&` band around zero was likely intended, so the first branch's second test is true for every value and the second branch's is redundant.

Rebased down from the original three sites: the triple shell arcs and the Banshee Boardwalk lamp range are already covered on master by #778 and #779.

Comment-only change, so the preprocessed output is byte-identical and matching is unaffected.


MegaMech:
oops, I didn't see this PR. But I pushed comments for the bugs except the first one iirc.

quarrel07:
No problem. I rebased this down to just the first one, the other two are covered 👍

Issue #715, opened 2025-05-13
Will the Japanese version be supported?


yyoossk:
Will the Japanese version be supported?

MegaMech:
Nah

PR #777, opened 2026-07-31
Add Japanese v1.1 as a fourth build target

quarrel07:
Adds revision 1.1 of the original December 1996 Japanese release, next to `us`, `eu.v10` and `eu.v11`. Drop in your own `baserom.jp.v11.z64`, run `make assets -j VERSION=jp.v11` then `make -j VERSION=jp.v11`, and it builds byte for byte.

| ROM Output      | Revision | SHA-1 Checksum                           |
|-----------------|----------|------------------------------------------|
| mk64.jp.v11.z64 | JPN 1.1  | 9f439457585146a4e1da7e1dd9104f7f94381688 |

**No game data is added.** 994 of the 1,067 `assets.json` entries gain a `jp.v11` offset, 12,408 group assets gain a `jp.v11` override, and there are 8 new Torch yamls under `yamls/jp.v11/`. `git ls-files` still shows exactly the three ROM-derived files this repo already had, which are the documentation images.

## What actually differs

Most of this is arrangement rather than content, but not all of it. The CPU code genuinely diverges.

**Gameplay and CPU behavior**

* **Yoshi Valley's forks are chosen differently.** JP 1.1 rolls `gPlayerPathIndex = random_int(4)` when a CPU crosses the finish line, in place of the `yoshi_valley_cpu_path` path-tracking that every other version uses, which runs per path point. The JP 1.1 code sits under one guard; the existing function is untouched.
* **The JP revisions give CPUs a strategy for double and triple mushrooms.** The two `case` labels are in the US and EU builds with empty bodies, so their CPUs fall through to generic behavior.
* **JP 1.1 does not have an audio-engine typo that US and JP 1.0 both carry.** `src/audio/external.c` reads `if ((x > 0.5f) || (x < 0.5f))` for US and JP 1.0, which is true for every value except exactly 0.5. JP 1.1 reads `< -0.5f`, as does EU. The US side is left as-is, since it is what the cartridge does.
* **JP 1.1 (and possibly JP 1.0) draws less scenery.** In 14 of the 20 courses it issues fewer `gsSPDisplayList` calls than the US and EU builds over the same sub-models: 1,742 fewer in total. Mario Raceway is the cleanest case at 1,996 against their 2,123, with every other command type matching in count. Same models, invoked less often. JP 1.0 has not been counted, so this may go back further than 1.1.

**Different data**

* **Luigi Raceway is the one course with different geometry**: 5,949 vertices against 5,936, and 284 of its 358 display list arrays differ. It also tops the scenery table above, so it is both the course they remodeled and the one they dressed most heavily afterward.
* **`gTLUTCactus` has no JP counterpart anywhere in the cart.** That palette is simply absent, which is why those five cactus textures are the one place this PR reads indices raw instead of converting a png.

**Same data, moved**

This is the category that made the work hard, because a wrong offset and wrong content look identical until you decompress and compare.

* **The CPU behavior tables are byte-identical to US and EU.** All 17, and the four target-speed tables. JP just keeps them at the *end* of the common segment rather than ahead of the textures, which is why every texture in that segment sits 0xA40 lower than it does there.
* **The ceremony segment moves the other way around.** Four podium path tables go from the end of the segment to just before `light1`, and the 736 bytes between them shift up by 0x2E0.

**Text, credits and leftovers**

* **JP 1.1 drops four Japanese debug strings and adds one.** `迷路に突入！` (entering the maze), `迷路から出た！`, `バイパス切り替え エラー!!!` and `OGAWA DEBUG DRAW` are in every other version but not JP 1.1, going with the Yoshi Valley path code above. In their place JP 1.1 has `チョコマウンテン踏み外し！！！`, belonging to a Choco Mountain recovery path new in that revision.
* **The credits table is dual-language in every version**, which is worth saying because it looks like a JP-only structure and is not. `gCreditsText` is 126 entries either way, English in the first 63 and Japanese in the second, so `gCreditsTextExtra` is not a second table: it is `gCreditsText + 0xFC`, the same table entered at its midpoint. 11 of the 126 entries differ. JP credits six voice actors where US and EU credit three and leave three empty strings in the slots the others came out of; JP says `nintendo e.a.d.` where they say `phil sandhop`; and `rare ltd.` where they have the lawyered `Donkey Kong 3-D Model Provided Courtesy of Rare U.K.`
* **Three course-name tables where US and EU have one**: romanized (`mario circuit`), Japanese, and Japanese with `ドンキージャングルパーク` shortened to `ドンキージャングル` to fit. One name of twenty differs between the last two.
* **JP 1.1 fixes a Japanese misspelling and keeps an English one.** Every other version carries an extra ン in `サンプリンング ボイス`, where JP 1.1 has `サンプリング`. Both JP revisions spell `charles martinee`, missing the final t, where US and EU have `charles martinet`.
* **JP 1.1 hardcodes `gRefreshRate = 60`** where US and JP 1.0 switch on `osTvType` for NTSC, PAL and MPAL.

## What is in the diff

40 commits, 250 files, +48,913 / -24,295.

| Area | Files | Lines |
|---|---|---|
| asset maps (json) | 49 | +13,527 / -13,361 |
| courses/ | 70 | +28,021 / -9,859 |
| yamls/ (Torch) | 8 | +3,350 |
| src, include, data, asm | 35 | +3,248 / -811 |
| .mk rules | 57 | +268 / -191 |
| tools/ | 20 | +131 / -8 |
| root files | 11 | +368 / -65 |

The asset map churn looks larger than it is. Nearly every one of those lines is a `jp.v11` override added inline to an existing entry, in the same idiom `assets.json` already used. No existing US offset changes value; 8 of the 1,026 entries gain a nested `meta.overrides` block, and that is the whole of the US side's exposure.

Five assets read `"type": "bin"` for JP rather than going through the normal image path: the kalimari cacti, for the palette reason above.

Source differences are `#ifdef VERSION_JP`, matching the existing `#ifdef VERSION_EU` style: 1,678 guard sites across 52 files. 9 of them read `#if defined(VERSION_EU) || defined(VERSION_JP)`, where EU and JP agree and US is the odd one out. That is also why this does not adopt oot's ordered version integers, which assume a linear lineage that mk64 does not have.

### luigi_raceway

The vertex table and the display lists both go in as separate files under a single `#ifdef VERSION_JP` (`course_vertices.jp.v11.inc.c` and `course_displaylists.jp.v11.inc.c`) rather than inline guards. Inline would need roughly 284 of them sitting inside array bodies, in files that today produce byte-perfect `us`, `eu.v10` and `eu.v11`. A separate file cannot regress those.

Both were produced by disassemblers that are checked against your own US files first: emit the US table from the US cart and the result matches the committed file character for character. So the JP output is in your format rather than a new one, and both files are clang-format clean.

### Why tools/generate_jp_course_names.py exists

JP shows three forms of each course name where US shows one: romanized, Japanese, and Japanese with one name shortened to fit. Torch's `mk64:metadata` factory writes one names table per config entry from the fixed `name` field, and emits twenty files from the same course metadata with every field required.

Three names tables that way means three copies of all twenty-one fully populated course yamls, differing by a single string each, re-emitting nineteen identical tables alongside. So the three tables come from a small json plus a generator instead. If the factory ever takes the names table's source field from its config entry, that json folds back into `yamls/courses` and nothing else changes. Happy to do it the other way if you would rather.

## Verification

Fresh clone, only the four cartridge dumps copied in, every version built and compared byte for byte against its own dump. Not just hashed. Run in both orders, because building JP first in a clean tree is what catches assets left behind by a previous version.

| Version | SHA-1 | `us` first | `jp.v11` first |
|---|---|---|---|
| `us` | `579c48e211ae952530ffc8738709f078d5dd215e` | matches | matches |
| `eu.v10` | `a729039453210b84f17019dda3f248d5888f7690` | matches | matches |
| `eu.v11` | `f6b5f519dd57ea59e9f013cc64816e9d273b2329` | matches | matches |
| `jp.v11` | `9f439457585146a4e1da7e1dd9104f7f94381688` | matches | matches |

`git status` is empty after all four in both orders, so nothing the build needs is untracked. `jp.v11` builds byte-correct as the first thing in a fresh clone, without `us` built first.

## Why the asset pipeline commits are here

Several commits make the asset pipeline version-aware rather than adding JP data. They are not separable from this PR: JP is the first version whose asset *bytes* differ rather than merely moving, so without them a `VERSION` switch silently compiles whatever the previous build extracted. It is also why the verification above is run in both orders.

One of them turns out to fix something on master. `assets: re-extract the group assets when the built version changes` adds `ASSET_VERSION := $(TORCH_VERSION)`, folding EU's asset extraction to US the way its Torch output already did. Take `230fe2ec1`, cherry-pick only the three macOS build fixes so it compiles on a Mac, then follow the documented `make assets && make VERSION=eu.v10` in a fresh clone: you get `c600563dc5abed8ed616bcab3beb0ba4bd3dc3ba` rather than `a729039453...`. None of the 12,521 group assets carries an EU offset, so the moment extraction actually runs for EU it reads the EU cart at US positions. A warm tree hides it, because a previous US build has already left the right files on disk.

## Notes for the review

Formatting was applied with `git-clang-format` against `230fe2ec1` using the pinned clang-format 14, so only lines this branch touched were considered. That came to a stray blank line and one continuation alignment. A bare `make format` would have rewritten 341 lines in files this PR barely touches, since the tree is not currently clean under its own pinned version.

Two files are deliberately left unformatted. `src/audio/external.c` would be re-indented to suit the JP branch, which nests a level deeper around a body both branches share, so the result is right for JP and wrong for US. `src/menu_items.c` is EUC-JP and clang-format counts bytes rather than columns, so its wrapping there is not meaningful. Both generated luigi files are clang-format clean.

CI is not wired up for `jp.v11`, since that needs an encrypted JP ROM and a matching secret only you can add. Happy to do it if you want it.

JP 1.0, the launch revision, is a separate and larger job: measured against two reference points it lands closer to the US delta than to the gap between the two EU revisions. Later PR.

## Depends on

#772 for the macOS build fixes, and #776 for the doxygen step. Two commits here are #772's and one is #776's; all three drop out as those merge.


src/racing/race_logic.c
```diff
@@ -788,7 +788,11 @@ void func_8028F588(void) {
 }
 
 void func_8028F8BC(void) {
+#ifdef VERSION_JP
+    D_802BA034 = (f32) (D_802BA034 - 0.017);
+#else
     D_802BA034 = (f32) (D_802BA034 - 0.017f);
+#endif
```
MegaMech, line 795: Any chance US matches without the f? `0.017`

src/cpu_vehicles_camera_path/cpu_item_strategy.inc.c
```diff
@@ -36,10 +36,20 @@ void cpu_decisions_branch_item(UNUSED s32 playerId, s16* branch, s32 itemId) {
...
             value = CPU_STRATEGY_ITEM_MUSHROOM;
             break;
+#ifdef VERSION_JP
+        // JP also gives CPUs a strategy for double/triple mushrooms
         case ITEM_DOUBLE_MUSHROOM:
+            value = CPU_STRATEGY_ITEM_DOUBLE_MUSHROOM;
             break;
         case ITEM_TRIPLE_MUSHROOM:
+            value = CPU_STRATEGY_ITEM_TRIPLE_MUSHROOM;
             break;
```
MegaMech: I feel like I'd rather have the ifdef in the cases here, just so the cases don't get duplicated...

But it probalby doens't really matter either way.

MegaMech (changes requested):

MegaMech:
Wow... lgtm

MegaMech:
Does US still build if you make clean and build US after building EU?

src/cpu_vehicles_camera_path/cpu_item_strategy.inc.c
quarrel07: Agreed, that reads better. Done in cbe47db4a: the case labels appear once and only the assignment is guarded. All four ROMs still byte-identical after it.

src/racing/race_logic.c
quarrel07: Afraid not, I tried it. Dropping the guard and building `us` with `0.017` shifts the ROM from `func_8028F8BC` onward. With the double literal the compiler takes the double path:
```
    cvt.d.s  $f6,$f4
    sub.d    $f10,$f6,$f8
    cvt.s.d  $f16,$f10
```
where the `us` cart stays single precision throughout. So the two carts genuinely compute that one differently, which is what the ifdef is recording.

quarrel07:
> Does US still build if you make clean and build US after building EU?

Yes. Tested it two ways just now:
```
    build eu.v10, then make clean, then build us  -> byte-identical
    rm -rf build entirely, then build us          -> byte-identical
```
Worth saying why that is not automatic: `make clean` only removes `build/`, not the extracted assets, so it is exactly the case where the previous version's art is still sitting on disk. It holds because of the assets/.version stamp added in "assets: re-extract the group assets when the built version changes". That is the same commit the description credits for EU, and on master today this is the scenario that goes wrong.

MegaMech:
Sorry, I meant JP not EU
I think `make distclean` is equivallent to your `rm -rf`

quarrel07:
> Sorry, I meant JP not EU I think `make distclean` is equivallent to your `rm -rf`

All good! Just ran that one in a single reused tree:

`build jp.v11, then gmake clean, then build us  ->  byte-identical`

Worth noting `gmake clean` with no `VERSION` only removes `build/us`, since `BUILD_DIR` is `build/$(VERSION)`, so `build/jp.v11` was still sitting there when US built, and the extracted assets on disk were still the JP ones. Came out byte-identical anyway, which is the `assets/.version` stamp doing its job.

> I think `make distclean` is equivallent to your `rm -rf`

The one difference is `distclean` also runs `extract_assets.py --clean`, so the extracted art goes too where `make clean` leaves it on disk.

MegaMech:
Lets remove the macos IDO Recomp binaries for now and figure something out later for that.

quarrel07:
Sounds good, I'll pull them out of #777 and #772.

queueRAM's ido-static-recomp submodule idea looks like the better fix to me anyway. I tried it here just now, make setup plus make VERSION=5.3 was about 8 seconds and mk64 builds a byte-matching US ROM off the result. Happy to do that as its own PR if it would help.

MegaMech:
Resolves https://github.com/n64decomp/mk64/issues/715

